PR #6, opened 2022-09-08
adds embedders as code functionality

jhoetter:


jhoetter:
as discussed in our roadmap [thread](https://github.com/code-kern-ai/refinery/discussions/81), we're going to postpone this feature for some versions

JWittmeyer:
over half a year old

PR #9, opened 2022-09-15
Create new attribute

lumburovskalina:
works with
- https://github.com/code-kern-ai/refinery-embedder/pull/14
- https://github.com/code-kern-ai/refinery-ac-exec-env/pull/1
- https://github.com/code-kern-ai/refinery-gateway/pull/42
- https://github.com/code-kern-ai/refinery-ui/pull/54
- https://github.com/code-kern-ai/refinery-tokenizer/pull/6
- https://github.com/code-kern-ai/refinery-submodule-model/pull/9


business_objects/attribute.py
```diff
@@ -50,27 +50,51 @@ def get_all_by_names(project_id: str, attribute_names: List[str]) -> List[Attrib
     )
 
 
-def get_all(project_id: str) -> List[Attribute]:
-    return session.query(Attribute).filter(Attribute.project_id == project_id).all()
+def get_all(
+    project_id: str, state_filter: Optional[List[str]] = None
+) -> List[Attribute]:
+    if state_filter is None:
```
jhoetter: caution, this is a bit misleading; as the user of this function, i would assume that setting `state_filter` to `None` will actually give me _all_ states.
another option would be to use the lists as the default argument (`state_filter: List[str] = [AttributeState.UPLOADED.value, AttributeState.USABLE.value]`).
generally, i would not recommend this, as the list is only created once as the function is defined (see https://docs.python-guide.org/writing/gotchas/), but in this case, it doesn't matter, as we never append to the default list.

just my two cents, handle however you like :)

business_objects/attribute.py
```diff
@@ -50,27 +50,51 @@ def get_all_by_names(project_id: str, attribute_names: List[str]) -> List[Attrib
...
+        query = query.filter(Attribute.state.in_(state_filter))
+    return query.all()
 
 
-def get_attribute_ids(project_id: str) -> Dict[str, str]:
-    attributes: List[Attribute] = get_all(project_id)
+def get_attribute_ids(
+    project_id: str, state_filter: Optional[List[str]] = None
+) -> Dict[str, str]:
+    if state_filter is None:
```
jhoetter: see above

business_objects/attribute.py
```diff
@@ -50,27 +50,51 @@ def get_all_by_names(project_id: str, attribute_names: List[str]) -> List[Attrib
...
-    text_attributes = (
-        session.query(Attribute)
-        .filter(Attribute.project_id == project_id, Attribute.data_type == "TEXT")
-        .all()
-    )
+def get_text_attributes(
+    project_id: str, state_filter: Optional[List[str]] = None
+) -> Dict[str, str]:
+    if state_filter is None:
+        state_filter = [AttributeState.UPLOADED.value, AttributeState.USABLE.value]
```
jhoetter: see above

business_objects/attribute.py
```diff
@@ -50,27 +50,51 @@ def get_all_by_names(project_id: str, attribute_names: List[str]) -> List[Attrib
...
+    text_attributes = query.all()
     return {att.name: str(att.id) for att in text_attributes}
 
 
-def get_all_ordered(project_id: str, order_asc: bool) -> List[Attribute]:
+def get_all_ordered(
+    project_id: str, order_asc: bool, state_filter: Optional[List[str]] = None
+) -> List[Attribute]:
+    if state_filter is None:
+        state_filter = [AttributeState.UPLOADED.value, AttributeState.USABLE.value]
```
jhoetter: repeating myself here :D

jhoetter (approved):
nice work :)

business_objects/attribute.py
FelixKirsch: Changed to usage of the list as default argument

PR #10, opened 2022-09-19
adds confidence chart

jhoetter:
use [gateway](https://github.com/code-kern-ai/refinery-gateway/pull/45) and [frontend](https://github.com/code-kern-ai/refinery-ui/pull/60) from this branch.

When you jump into the clickbait project, you should see a confidence distribution chart on the overview page

### DIFF
--- a/business_objects/project.py
+++ b/business_objects/project.py
@@ -3,7 +3,13 @@ from typing import List, Optional, Any, Dict, Union
 from . import general
 from .. import enums
 from ..session import session
-from ..models import Project
+from ..models import (
+    DataSliceRecordAssociation,
+    LabelingTask,
+    LabelingTaskLabel,
+    Project,
+    RecordLabelAssociation,
+)
 
 
 def get(project_id: str) -> Project:
@@ -58,6 +64,38 @@ def get_label_distribution(
     )
     if values:
         return values[0]
+
+
+def get_confidence_distribution(
+    project_id: str,
+    labeling_task_id: str,
+    data_slice_id: str = None,
+    num_samples: int = None,
+) -> List[float]:
+    query_filter = session.query(RecordLabelAssociation.confidence).filter(
+        RecordLabelAssociation.project_id == project_id,
+        LabelingTask.id == labeling_task_id,
+        LabelingTask.id == LabelingTaskLabel.labeling_task_id,
+        RecordLabelAssociation.labeling_task_label_id == LabelingTaskLabel.id,
+        RecordLabelAssociation.source_type == enums.LabelSource.WEAK_SUPERVISION.value,
+    )
+
+    if data_slice_id is not None:
+        query_filter = query_filter.filter(
+            DataSliceRecordAssociation.data_slice_id == data_slice_id,
+            DataSliceRecordAssociation.record_id == RecordLabelAssociation.record_id,
+            DataSliceRecordAssociation.project_id == project_id,
+        )
+
+    if num_samples is not None:
+        query_filter = query_filter.limit(num_samples)
+        confidence_scores = [confidence for confidence, in (query_filter.all())]
+        confidence_scores = sorted(confidence_scores)
+    else:
+        query_filter = query_filter.order_by(RecordLabelAssociation.confidence.asc())
+        confidence_scores = [confidence for confidence, in (query_filter.all())]
+
+    return confidence_scores
 
 
 def get_confusion_matrix(
@@ -130,7 +168,7 @@ def create(
     created_by: str,
     created_at: Optional[str] = None,
     with_commit: bool = False,
-    status: enums.ProjectStatus = enums.ProjectStatus.INIT_UPLOAD
+    status: enums.ProjectStatus = enums.ProjectStatus.INIT_UPLOAD,
 ) -> Project:
     project: Project = Project(
         name=name,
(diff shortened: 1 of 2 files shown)


business_objects/project.py
```diff
@@ -60,6 +66,38 @@ def get_label_distribution(
...
+    data_slice_id: str = None,
+    num_samples: int = None,
+) -> List[float]:
+    query_filter = session.query(RecordLabelAssociation.confidence).filter(
+        RecordLabelAssociation.project_id == project_id,
+        LabelingTask.id == labeling_task_id,
+        LabelingTask.id == LabelingTaskLabel.labeling_task_id,
+        RecordLabelAssociation.labeling_task_label_id == LabelingTaskLabel.id,
+        RecordLabelAssociation.source_type == enums.LabelSource.WEAK_SUPERVISION.value,
+    )
```
JWittmeyer: newly added featues use the join option of SQL alchemy to ensure we use a real join and not a cross apply with filter.
Compare `get_labeling_tasks_by_selected_sources` for reference

business_objects/project.py
```diff
@@ -60,6 +66,38 @@ def get_label_distribution(
...
+
+    if data_slice_id is not None:
+        query_filter = query_filter.filter(
+            DataSliceRecordAssociation.data_slice_id == data_slice_id,
+            DataSliceRecordAssociation.record_id == RecordLabelAssociation.record_id,
+            DataSliceRecordAssociation.project_id == project_id,
+        )
+
+    if num_samples is not None:
+        query_filter = query_filter.limit(num_samples)
```
JWittmeyer: i think for a limit we should also at least order by random to not create a pseudo-deterministic state

business_objects/project.py
```diff
@@ -60,6 +66,38 @@ def get_label_distribution(
         return values[0]
 
 
+def get_confidence_distribution(
+    project_id: str,
+    labeling_task_id: str,
+    data_slice_id: str = None,
+    num_samples: int = None,
```
JWittmeyer: type optoinal

business_objects/project.py
jhoetter: great tip, implemented it :)

business_objects/project.py
jhoetter: adding a random order would result in pulling different data whenever you reload the overview page -> can't compare to previous state anymore. picking the first N as the random sample is ok in this case.

business_objects/project.py
jhoetter: done

business_objects/project.py
JWittmeyer: the first N are also not deterministic that s why I wouldn't want the pseudo determinisitc state that looks like a comparable state but isn't --> resolve if still irrelevant :)

jhoetter: understood. maybe not completely irrelevant, and interesting for later, but good for now :)

business_objects/project.py
```diff
@@ -60,6 +66,48 @@ def get_label_distribution(
...
+    project_id: str,
+    labeling_task_id: str,
+    data_slice_id: Optional[str] = None,
+    num_samples: Optional[int] = None,
+) -> List[float]:
+    query_filter = (
+        session.query(RecordLabelAssociation.confidence)
+        .join(
+            LabelingTaskLabel,
+            RecordLabelAssociation.labeling_task_label_id == LabelingTaskLabel.id,
```
JWittmeyer: please with complete join criteria (project_id) without it, it should still work (since the uuid is unique) however with the full join criteria the db optimizer should be able to reduce the data checked for the join significantly (thats why we indexed every fkey :) )

business_objects/project.py
```diff
@@ -60,6 +66,48 @@ def get_label_distribution(
...
+        )
+        .join(
+            LabelingTask,
+            LabelingTask.id == LabelingTaskLabel.labeling_task_id,
```
JWittmeyer: project_id missing

business_objects/project.py
```diff
@@ -60,6 +66,48 @@ def get_label_distribution(
...
+            RecordLabelAssociation.project_id == project_id,
+            LabelingTask.id == labeling_task_id,
+            RecordLabelAssociation.source_type
+            == enums.LabelSource.WEAK_SUPERVISION.value,
+        )
+    )
+
+    if data_slice_id is not None:
+        query_filter = query_filter.filter(
+            DataSliceRecordAssociation.data_slice_id == data_slice_id,
```
JWittmeyer, line 107: join missing

business_objects/project.py
jhoetter: implemented

business_objects/project.py
jhoetter: done

business_objects/project.py
jhoetter: done

business_objects/project.py
```diff
@@ -60,6 +66,54 @@ def get_label_distribution(
...
+        .join(
+            LabelingTask,
+            (LabelingTask.id == LabelingTaskLabel.labeling_task_id)
+            & (LabelingTask.project_id == LabelingTaskLabel.project_id),
+        )
+        .filter(
+            LabelingTask.id == labeling_task_id,
+            RecordLabelAssociation.source_type
+            == enums.LabelSource.WEAK_SUPERVISION.value,
+        )
```
FelixKirsch:
```suggestion
        .filter(
            RecordLabelAssociation.project_id == project_id,
            LabelingTask.id == labeling_task_id,
            RecordLabelAssociation.source_type
            == enums.LabelSource.WEAK_SUPERVISION.value,
        )
```
Filter by project_id, I guess this should do it

FelixKirsch (approved):

PR #11, opened 2022-09-20
access all manual token of a record

jhoetter:
together with [frontend](https://github.com/code-kern-ai/refinery-ui/pull/57) and [gateway](https://github.com/code-kern-ai/refinery-gateway/pull/46)

### DIFF
--- a/business_objects/record_label_association.py
+++ b/business_objects/record_label_association.py
@@ -93,18 +93,26 @@ def get_tokens(project_id: str) -> List[Any]:
 
 
 def get_manual_tokens_by_record_id(
+    project_id: str,
     record_id: str,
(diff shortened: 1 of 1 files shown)


business_objects/record_label_association.py
```diff
@@ -92,6 +92,24 @@ def get_tokens(project_id: str) -> List[Any]:
     return general.execute_all(query)
 
 
+def get_manual_tokens_by_record_id(
+    record_id: str,
```
FelixKirsch, line 97: add project_id to arguments and use it to filter to improve performance

jhoetter: done

PR #19, opened 2022-10-18
First non text column for comments on records

lumburovskalina:
Steps to implement:
1. Gets the first non text column
2. Display the property with the data 

Issue:
https://github.com/code-kern-ai/refinery-ui/issues/72 

PRs: 
1. https://github.com/code-kern-ai/refinery-ui/pull/73
2. https://github.com/code-kern-ai/refinery-submodule-model/pull/19
3. https://github.com/code-kern-ai/refinery-gateway/pull/70


business_objects/comments.py
```diff
@@ -1,4 +1,5 @@
 from datetime import datetime
+from submodules.model.business_objects import record
 from submodules.model.models import CommentData
```
JWittmeyer: i think these should be relative paths (the CommentData one as well) since submodules don't know where they will be located

business_objects/record.py
```diff
@@ -571,3 +571,14 @@ def __get_order_by(project_id: str, first_x: int = 3) -> str:
     if order != "":
         order = "ORDER BY " + order
     return order
+
+
+def get_first_no_text_column(project_id: str, xfkey: str) -> str:
```
JWittmeyer: name of the function is imo misleading (I first thought this only selects the no text column)

however, it selects the record data and the attribute name of the first non-text attribute.

business_objects/comments.py
```diff
@@ -80,7 +81,17 @@ def get_add_info_category(
...
         table_name = category.get_table_name()
-        name_col = category.get_name_col()
+
+        if category == enums.CommentCategory.RECORD:
+            columns = record.get_first_no_text_column(project_id, xfkey)
+            data = columns[0]
+            no_text_column = columns[1]
+            name_col = (
+                "'" + str(no_text_column) + ": " + str(data[no_text_column]) + "'"
+            )
```
JWittmeyer: i really like the idea of using a constant as a name col. to optimize a bit I would grab the value from the database directly (ill list the query further down)
```suggestion
            name_col = record.get_first_no_text_value(project_id,xfkey)
```

business_objects/record.py
```diff
@@ -571,3 +571,14 @@ def __get_order_by(project_id: str, first_x: int = 3) -> str:
...
+    query = f"""
+    SELECT data, a.name
+    FROM attribute a
+    	join record r on r.project_id = a.project_id
+    WHERE a.project_id = '{project_id}' AND a.data_type != 'TEXT' AND r.id = '{xfkey}'
+    limit 1
+    """
```
JWittmeyer:
```suggestion
def get_first_no_text_column(project_id: str, record_id: str) -> str:
    query = f"""
    SELECT '''' || x.name || ': ' || (r.data ->>x.name) || '''' AS name_col
    FROM record r,
    (
	    SELECT a.name
	    FROM attribute a 
	    WHERE data_type NOT IN('{enums.DataTypes.TEXT.value}' , '{enums.DataTypes.CATEGORY.value}')
		    AND a.project_id = '{project_id}'
	    LIMIT 1 
     )x
    WHERE r.project_id = '{project_id}' AND r.id = '{record_id}'

    """
```

business_objects/record.py
```diff
@@ -571,3 +571,19 @@ def __get_order_by(project_id: str, first_x: int = 3) -> str:
...
+    SELECT '''' || x.name || ': ' || (r.data ->>x.name) || '''' AS name_col
+    FROM record r,
+    (
+	    SELECT a.name
+	    FROM attribute a 
+	    WHERE data_type NOT IN('{enums.DataTypes.TEXT.value}' , '{enums.DataTypes.CATEGORY.value}')
+		    AND a.project_id = '{project_id}'
+	    LIMIT 1 
+     )x
+    WHERE r.project_id = '{project_id}' AND r.id = '{record_id}'
```
JWittmeyer: indentation is now a little bit funky because of the suggestion thing I think. maybe (uses tabs instead of common spaces)

JWittmeyer (approved):

PR #22, opened 2022-11-08
Label studio import

SirDegraf:
Submodule PR: https://github.com/code-kern-ai/refinery-submodule-model/pull/22
Gateway PR: https://github.com/code-kern-ai/refinery-gateway/pull/80
frontend PR: https://github.com/code-kern-ai/refinery-ui/pull/81

### DIFF
--- a/business_objects/labeling_task.py
+++ b/business_objects/labeling_task.py
@@ -223,14 +223,14 @@ def get_labeling_task_by_source_id(source_id: str) -> LabelingTask:
 
 def get_labeling_task_name_by_label_id(project_id: str) -> Dict:
     results: List = general.execute_all(
-        f"""SELECT ltl.id, lt.name
+        f"""SELECT ltl.id::TEXT, lt.name
         FROM labeling_task_label ltl 
-        JOIN labeling_task lt 
-        ON ltl.labeling_task_id = lt.id
+        INNER JOIN labeling_task lt 
+            ON ltl.project_id = lt.project_id AND ltl.labeling_task_id = lt.id
         WHERE lt.project_id = '{project_id}'
         """
     )
-    return {str(result[0]): result[1] for result in results}
+    return {result[0]: result[1] for result in results}
 
 
 def create(
--- a/business_objects/labeling_task.py
+++ b/business_objects/labeling_task.py
@@ -221,6 +221,18 @@ def get_labeling_task_by_source_id(source_id: str) -> LabelingTask:
     )
 
 
+def get_labeling_task_name_by_label_id(project_id: str) -> Dict:
+    results: List = general.execute_all(
+        f"""SELECT ltl.id, lt.name
+        FROM labeling_task_label ltl 
+        JOIN labeling_task lt 
+        ON ltl.labeling_task_id = lt.id
+        WHERE lt.project_id = '{project_id}'
+        """
+    )
+    return {str(result[0]): result[1] for result in results}
+
+
 def create(
     project_id: str,
     attribute_id: str,
--- a/business_objects/record_label_association.py
+++ b/business_objects/record_label_association.py
@@ -199,24 +199,16 @@ def get_all_classifications_for_information_source(
     )
 
 
-def get_existing_record_label_associations(
+def get_existing_manual_record_label_associations(
     project_id: str, record_ids: List[str], user_ids: List[str], label_ids: List[str]
 ) -> List[RecordLabelAssociation]:
     return (
         session.query(RecordLabelAssociation)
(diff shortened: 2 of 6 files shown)


business_objects/labeling_task.py
```diff
@@ -221,6 +221,18 @@ def get_labeling_task_by_source_id(source_id: str) -> LabelingTask:
...
+def get_labeling_task_name_by_label_id(project_id: str) -> Dict:
+    results: List = general.execute_all(
+        f"""SELECT ltl.id, lt.name
+        FROM labeling_task_label ltl 
+        JOIN labeling_task lt 
+        ON ltl.labeling_task_id = lt.id
+        WHERE lt.project_id = '{project_id}'
+        """
+    )
+    return {str(result[0]): result[1] for result in results}
```
JWittmeyer: str cast can be done in SQL with `ltl.id::TEXT`

business_objects/labeling_task.py
```diff
@@ -221,6 +221,18 @@ def get_labeling_task_by_source_id(source_id: str) -> LabelingTask:
...
 
 
+def get_labeling_task_name_by_label_id(project_id: str) -> Dict:
+    results: List = general.execute_all(
+        f"""SELECT ltl.id, lt.name
+        FROM labeling_task_label ltl 
+        JOIN labeling_task lt 
+        ON ltl.labeling_task_id = lt.id
+        WHERE lt.project_id = '{project_id}'
+        """
```
JWittmeyer: doesn't match our style guide & project_id join missing
```suggestion
        f"""SELECT ltl.id::TEXT, lt.name
        FROM labeling_task_label ltl 
        INNER JOIN labeling_task lt 
            ON ltl.project_id = lt.project_id AND ltl.labeling_task_id = lt.id
        WHERE lt.project_id = '{project_id}'
        """
```

business_objects/record_label_association.py
```diff
@@ -199,6 +199,28 @@ def get_all_classifications_for_information_source(
...
+    return (
+        session.query(RecordLabelAssociation)
+        .filter(RecordLabelAssociation.project_id == project_id)
+        .filter(RecordLabelAssociation.record_id.in_(record_ids))
+        .filter(
+            RecordLabelAssociation.created_by.in_(
+                [
+                    user_id
+                    for user_id in user_ids
+                    if user_id != enums.RecordImportMappingValues.IGNORE.value
```
JWittmeyer: this feels like a very specific exclude for an unspecific case.
if we give a list of user ids then I would expect that the list isn't further processed. Maybe remove IGNORE beforehand

business_objects/record_label_association.py
```diff
@@ -199,6 +199,28 @@ def get_all_classifications_for_information_source(
     )
 
 
+def get_existing_record_label_associations(
```
JWittmeyer: name suggests all but we only collect manual rlas

business_objects/record_label_association.py
```diff
@@ -342,6 +364,28 @@ def create_token_objects(
...
+                    project_id=project_id,
+                    record_id=record_id,
+                    labeling_task_label_id=label_id,
+                    source_type=enums.LabelSource.MANUAL.value,
+                    return_type=enums.InformationSourceReturnType.RETURN.value,
+                    created_by=user_id,
+                )
+                for label_id in tasks_dict.values()
+            ]
+            create_list.extend(rlas)
```
JWittmeyer, line 377: extending a list often will run into performance issues, especially once they reach a certain size since space needs to be allocated beforehand and then copied from a to b.

if think if we chunk the data beforehand it's not super dramatic however with currently unlimited amounts of data not optimal

business_objects/labeling_task.py
SirDegraf: changed

business_objects/labeling_task.py
SirDegraf: changed

business_objects/record_label_association.py
SirDegraf: renamed

business_objects/record_label_association.py
SirDegraf: changed so the filtering of IGNORE is done before calling of method

business_objects/record_label_association.py
SirDegraf: understood. Since it is now done in chunks i would leave it as it is but complelty valid point

JWittmeyer: works for me

JWittmeyer (approved):

PR #30, opened 2022-12-14
Tokenizer refactoring and attribute tokenization introduction

SirDegraf:
Tokenizer: https://github.com/code-kern-ai/refinery-tokenizer/pull/17
Submodule Model: https://github.com/code-kern-ai/refinery-submodule-model/pull/30
Gateway: https://github.com/code-kern-ai/refinery-gateway/pull/96

### DIFF
--- a/business_objects/attribute.py
+++ b/business_objects/attribute.py
@@ -92,7 +92,7 @@ def get_text_attributes(
     )
     if state_filter:
         query = query.filter(Attribute.state.in_(state_filter))
-    text_attributes = query.all()
+    text_attributes = query.order_by(Attribute.relative_position.asc()).all()
     return {att.name: str(att.id) for att in text_attributes}
 
 
--- a/business_objects/general.py
+++ b/business_objects/general.py
@@ -53,7 +53,7 @@ def flush_or_commit(commit: bool = False) -> None:
         session.flush()
 
 
-def execute(sql: str, *args) -> Any:
+def execute(sql: Any, *args) -> Any:
     return session.execute(sql, *args)
 
 
--- a/business_objects/general.py
+++ b/business_objects/general.py
@@ -53,8 +53,8 @@ def flush_or_commit(commit: bool = False) -> None:
         session.flush()
 
 
-def execute(sql: str) -> Any:
-    return session.execute(sql)
+def execute(sql: str, *args) -> Any:
+    return session.execute(sql, *args)
 
 
 def execute_all(sql: str) -> List[Any]:
--- a/business_objects/record.py
+++ b/business_objects/record.py
@@ -103,12 +103,14 @@ def get_attribute_data_with_doc_bins_of_records(
     project_id: str, attribute_name: str
 ) -> List[Any]:
     query = f"""
-    SELECT rt.id, r."data" ->> '{attribute_name}' "attribute_data", rt.bytes
-    from record r
-    JOIN record_tokenized rt
-    ON  r.id = rt.record_id
+    SELECT 
+        rt.id, 
+        r."data" ->> '{attribute_name}' attribute_data, 
+        rt.bytes
+    FROM record r
+    INNER JOIN record_tokenized rt
+        ON  r.project_id = rt.project_id AND r.id = rt.record_id
     WHERE r.project_id = '{project_id}'
-    AND rt.project_id = '{project_id}'
     """
     return general.execute_all(query)
 
--- a/business_objects/record.py
+++ b/business_objects/record.py
@@ -2,6 +2,8 @@ from __future__ import with_statement
 from typing import List, Dict, Any, Tuple
 from sqlalchemy import cast, Text
 from sqlalchemy.orm.attributes import flag_modified
+from sqlalchemy.sql.expression import bindparam
+from sqlalchemy import update
 
 from . import attribute, general
 from .. import models, enums
@@ -11,6 +13,7 @@ from ..models import (
     LabelingTaskLabel,
     RecordAttributeTokenStatistics,
     Attribute,
+    RecordTokenized,
 )
 from ..session import session
 
@@ -88,6 +91,38 @@ def get_ids_of_manual_records_by_labeling_task(
     )
 
 
+def get_attribute_data_with_doc_bins_of_records(project_id, attribute_name):
+    query = f"""
+    SELECT rt.id, r."data" ->> '{attribute_name}' "attribute_data", rt.bytes
+    from record r
+    JOIN record_tokenized rt
+    ON  r.id = rt.record_id
+    WHERE r.project_id = '{project_id}'
+    AND rt.project_id = '{project_id}'
+    """
+    return general.execute_all(query)
+
+
+def update_bytes_of_record_tokenized(values: List[Dict[str, Any]]) -> None:
+    query = (
+        update(RecordTokenized)
+        .where(RecordTokenized.id == bindparam("_id"))
+        .values({"bytes": bindparam("bytes")})
+    )
+    general.execute(query, values)
+    general.flush()
+
+
+def update_columns_of_tokenized_records(rt_ids: List[str], attribute_name: str):
+    query = f"""
+    UPDATE record_tokenized 
+    SET columns = array_append(columns, '{attribute_name}')
+    WHERE id IN {rt_ids}
+    """
+    general.execute(query)
+    general.flush()
+
+
 def get_missing_rats_records(
     project_id: str, limit: int, attribute_id: str
 ) -> List[Any]:
@@ -161,6 +196,16 @@ def get_count_scale_uploaded(project_id: str) -> int:
         session.query(models.Record)
         .filter(
             models.Record.category == enums.RecordCategory.SCALE.value,
+            models.Record.project_id == project_id,
+        )
+        .count()
+    )
+
+
+def get_count_all_records(project_id: str) -> int:
+    return (
+        session.query(models.Record)
+        .filter(
             models.Record.project_id == project_id,
         )
         .count()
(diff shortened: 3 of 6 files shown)


business_objects/record.py
```diff
@@ -88,6 +91,48 @@ def get_ids_of_manual_records_by_labeling_task(
     )
 
 
+def count_records_without_tokenization(project_id: str) -> int:
+    query = f"""
+    SELECT count(r.id) c
+    FROM ({get_records_without_tokenization(project_id, 0, True)}) r
+    """
+    return general.execute(query).first().c
```
JWittmeyer: i think this would be execute_first?

business_objects/record.py
```diff
@@ -88,6 +91,48 @@ def get_ids_of_manual_records_by_labeling_task(
...
+def get_attribute_data_with_doc_bins_of_records(
+    project_id: str, attribute_name: str
+) -> List[Any]:
+    query = f"""
+    SELECT rt.id, r."data" ->> '{attribute_name}' "attribute_data", rt.bytes
+    from record r
+    JOIN record_tokenized rt
+    ON  r.id = rt.record_id
+    WHERE r.project_id = '{project_id}'
+    AND rt.project_id = '{project_id}'
```
JWittmeyer: Styleguide issues

```suggestion
    SELECT 
        rt.id, 
        r."data" ->> '{attribute_name}' attribute_data, 
        rt.bytes
    FROM record r
    INNER JOIN record_tokenized rt
        ON  r.project_id = rt.project_id AND r.id = rt.record_id
    WHERE r.project_id = '{project_id}'
```

JWittmeyer: also this will only result in values if the record already has something in the record_tokenized table. Is this wanted?

business_objects/record.py
```diff
@@ -88,6 +91,48 @@ def get_ids_of_manual_records_by_labeling_task(
...
+    """
+    return general.execute_all(query)
+
+
+def update_bytes_of_record_tokenized(values: List[Dict[str, Any]]) -> None:
+    query = (
+        update(RecordTokenized)
+        .where(RecordTokenized.id == bindparam("_id"))
```
JWittmeyer: i think we should still include the project id for indexing purposes

business_objects/record.py
```diff
@@ -88,6 +91,48 @@ def get_ids_of_manual_records_by_labeling_task(
...
+    )
+    general.execute(query, values)
+    general.flush()
+
+
+def update_columns_of_tokenized_records(rt_ids: List[str], attribute_name: str) -> None:
+    query = f"""
+    UPDATE record_tokenized 
+    SET columns = array_append(columns, '{attribute_name}')
+    WHERE id IN {rt_ids}
```
JWittmeyer, line 137: does this work? The ids are a list which in turn should result in something like ['id1','id2'] --> SQL needs () imo.

Usually we do something like `rla_id_filter = "'" + "', '".join(rla_ids) + "'"` + `AND rla.id IN ({rla_id_filter})`

business_objects/record.py
```diff
@@ -112,17 +157,45 @@ def get_missing_rats_records(
...
+        AND r.project_id = rt.project_id
         AND rt.project_id = '{project_id}'
-    WHERE r.project_id = '{project_id}' 
-    AND rt.id IS NULL    
+    WHERE r.project_id = '{project_id}'
+    AND rt.id IS NULL
+
     """
+    if query_only:
+        return query
```
JWittmeyer, line 185: shouldn't the query include the limit even if it's just a subquery?

business_objects/record.py
```diff
@@ -112,17 +157,45 @@ def get_missing_rats_records(
...
+        AND r.project_id = rt.project_id
+        AND rt.project_id = '{project_id}'
+    WHERE r.project_id = '{project_id}'
+    AND (
+        NOT rt."columns" @> '{attribute_names_string}'
+        OR rt."columns" IS NULL
+    )
+    """
+    if query_only:
+        return query
```
JWittmeyer, line 211: see above

business_objects/record.py
```diff
@@ -309,19 +392,14 @@ def count_missing_rats_records(project_id: str, attribute_id: str) -> int:
     return result.c
 
 
-def count_missing_tokenized_records(project_id: str) -> int:
+def count_missing_tokenized_records(project_id: str) -> List[Any]:
```
JWittmeyer: why does this now return an array (at least going by the function signature)

business_objects/record.py
```diff
@@ -309,19 +392,14 @@ def count_missing_rats_records(project_id: str, attribute_id: str) -> int:
...
     query = f"""
-    SELECT COUNT(*) c
-    FROM record r
-    LEFT JOIN record_tokenized rt
-        ON r.id = rt.record_id 
-        AND r.project_id = rt.project_id 
-        AND rt.project_id = '{project_id}'
-    WHERE r.project_id = '{project_id}' 
-    AND rt.id IS NULL    
+    SELECT COUNT(record_query.id)
```
JWittmeyer: COUNT of a column should only be used for specific cases where we explicitly want to count that column.
This is due to a table scan for NULL values of the column since they wont be counted towards the result.

So since we want to count all entrys (and pkey can't be null) I'd go with COUNT(*).

business_objects/record.py
```diff
@@ -112,17 +157,45 @@ def get_missing_rats_records(
...
         AND rt.project_id = '{project_id}'
-    WHERE r.project_id = '{project_id}' 
-    AND rt.id IS NULL    
+    WHERE r.project_id = '{project_id}'
+    AND rt.id IS NULL
+
     """
+    if query_only:
+        return query
+    if limit > 0:
```
JWittmeyer, line 186: the default value is 0, however, the function is also used with None. This will still have the same result however < 0 would result in some [strange behavior on None types ](https://stackoverflow.com/a/7383866/19801189) (wouldn't make sense context-wise but to prevent copy-paste issues I would check for none or ensure the function isn't used with None)

business_objects/record.py
SirDegraf: Changed to execute first

business_objects/record.py
SirDegraf: set a default of 0 for limit

business_objects/record.py
SirDegraf: change to count(*)

business_objects/record.py
SirDegraf: Changed back to int

business_objects/record.py
SirDegraf: In theory it would be fine but the subquery is used to count all the records for this condition

business_objects/record.py
SirDegraf: Yeah you are right it would be a problem but rt_ids is a string parsed to fit the expected format

business_objects/record.py
SirDegraf: In theory i would do it but for this query i would leave it as it is since only checking for the project_id as second where condition just created errors for me

business_objects/attribute.py
```diff
@@ -92,7 +92,7 @@ def get_text_attributes(
     )
     if state_filter:
         query = query.filter(Attribute.state.in_(state_filter))
-    text_attributes = query.all()
+    text_attributes = query.order_by(Attribute.relative_position.asc()).all()
```
JWittmeyer, line 95: not sure why we order and then put it in a dictionary (most programming languages define dictionaries as unordered, python since 3.7 has it ordered however so it will impact an iteration over the dict to some extent).

If it's for iteration purposes then we can just keep it, just wanted to be sure that it's intended

business_objects/general.py
```diff
@@ -53,8 +53,8 @@ def flush_or_commit(commit: bool = False) -> None:
         session.flush()
 
 
-def execute(sql: str) -> Any:
-    return session.execute(sql)
+def execute(sql: str, *args) -> Any:
+    return session.execute(sql, *args)
```
JWittmeyer: usually we use the execute method with a string object.
If I found it correctly the *arg were introduced for the update thing which doesn't use a string anymore. Maybe we should use a different method for this so there is no confusion in the future?

e.g.:`execute_query(query:<type of query object>, *args)`

business_objects/record.py
JWittmeyer: not sure what you mean, i tested with:
```
project_id = "9221e93a-e8d1-4fcb-b4a7-967cc7d13b21"
    query = (
        update(RecordTokenized)
        .where(
            RecordTokenized.id == bindparam("_id"),
            RecordTokenized.project_id == project_id,
        )
        .values({"bytes": bindparam("bytes")})
    )
```

and didn't receive any errors (project id of course would be a parameter instead) -> maybe I'm missing something though

business_objects/record.py
JWittmeyer: then we should change the type in the function header so people don't get irritiated

business_objects/record.py
SirDegraf: strange, maybe i am missing something, but for me i think i got problems here, but anyways it was the not working version from before, so yes lets take the project id in here for indexing if it is just the code shown above (project id as parameter ofc)

JWittmeyer (approved):

business_objects/general.py
SirDegraf: The args could become handy for other executions, too. Of course i can add a dedicated function for this but the args are per default just optional and can be ignored if not needed, in my perspective it just makes the method a bit more powerful

business_objects/record.py
SirDegraf: resolved

business_objects/record.py
SirDegraf: resolved

business_objects/attribute.py
SirDegraf: yes it is on purpose

business_objects/general.py
SirDegraf: Changed type of sql from str to Any

PR #31, opened 2023-02-03
Model Changes For Gates Update Endpoints

FelixKirsch:
see:
https://github.com/code-kern-ai/refinery-gateway/pull/103

### DIFF
--- a/business_objects/embedding.py
+++ b/business_objects/embedding.py
@@ -48,6 +48,10 @@ def get_embedding_id_and_type(project_id: str, embedding_name: str) -> Any:
     )
 
 
+def get_project_embeddings(project_id: str) -> List[Embedding]:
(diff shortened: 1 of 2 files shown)


business_objects/embedding.py
```diff
@@ -48,6 +48,10 @@ def get_embedding_id_and_type(project_id: str, embedding_name: str) -> Any:
     )
 
 
+def get_project_embeddings(project_id: str) -> List[Embedding]:
+    return session.query(Embedding).filter(Embedding.project_id == project_id).all()
```
JWittmeyer, line 52: might be mistaken but wouldn't get_finished_embeddings be sufficient? it has the project id and a state check. Error embedings or the ones currently being processed aren't necessary fo the update case anyway right?

JWittmeyer (approved):

PR #33, opened 2023-02-08
Gates Integration Improvements

FelixKirsch:
use with:

https://github.com/code-kern-ai/refinery-gateway/pull/111
https://github.com/code-kern-ai/refinery-submodule-model/pull/33

### DIFF
--- a/business_objects/payload.py
+++ b/business_objects/payload.py
@@ -3,7 +3,7 @@ from typing import List, Optional
 
 from . import general
 from .. import enums
-from ..models import InformationSource, InformationSourcePayload
+from ..models import InformationSourcePayload, InformationSource
 from ..session import session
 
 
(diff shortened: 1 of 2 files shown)


business_objects/payload.py
```diff
@@ -18,6 +18,21 @@ def get(project_id: str, payload_id: str) -> InformationSourcePayload:
...
+def get_first_running_active_learner(project_id: str) -> InformationSourcePayload:
+    return (
+        session.query(InformationSourcePayload, InformationSource)
+        .filter(
+            InformationSource.project_id == project_id,
+            InformationSource.type == enums.InformationSourceType.ACTIVE_LEARNING.value,
+            InformationSource.id == InformationSourcePayload.source_id,
+            InformationSourcePayload.project_id == project_id,
+            InformationSourcePayload.state != enums.PayloadState.FINISHED.value,
+            InformationSourcePayload.state != enums.PayloadState.FAILED.value,
```
JWittmeyer: why not = CREATED?

business_objects/payload.py
```diff
@@ -18,6 +18,21 @@ def get(project_id: str, payload_id: str) -> InformationSourcePayload:
...
+        )
```
JWittmeyer: this uses an approach we dont want to use anymore. With this approach, we are creating a cross apply with filter (which does pretty much what you want but isn't optimized / more error-prone). Also I think you only want to query one table since the return value is only Payload.

```suggestion
        session.query(InformationSourcePayload)
        .join(
            InformationSource,
            (InformationSourcePayload.source_id== InformationSource.id)
            & (InformationSourcePayload.project_id == InformationSource.project_id),
        )
        .filter(
            InformationSourcePayload.project_id == project_id,
            InformationSource.type == enums.InformationSourceType.ACTIVE_LEARNING.value,
            InformationSourcePayload.state == enums.PayloadState.CREATED.value,
        )
```
Use this instead :) 

(wrote it blind so might have a syntax error :D )

JWittmeyer (approved):

PR #36, opened 2023-02-21
Admin features

### LINKED ISSUE #239: Admin features
**Is your feature request related to a problem? Please describe.**
Adding new page for admin messages for notifying users, adding last interaction for users and check if something is running (ex. heuristics..).

**Describe the solution you'd like**
Admin dashboard:
Admin messages page: create new page, where the user can create new admin messages, archive them manually with an archive reason, and see the data for those messages (who and when created the message, who and when archived the messages, type of the message and if the message is still active or not). Add the automatically archiving of the messages. 
Extending users page: add field in the users page that is displaying the last interaction of the user. Add filter than can display the active users in the last X minutes (or/and a button for showing all active users). 
Check if something is running page: add new page that shows which processes are still running (ex heuristics...) and show more data about those.
Platform:
Implement websocket notifications on refinery, workflow and gates and add the active messages on the platform. The messages should be removed once the archive date is not valid.

SirDegraf:
Gateway https://github.com/code-kern-ai/refinery-gateway/pull/119
Model submodule https://github.com/code-kern-ai/refinery-submodule-model/pull/36
Admin dashboard: https://github.com/code-kern-ai/admin-dashboard/pull/62
Refinery UI: https://github.com/code-kern-ai/refinery-ui/pull/125
Updater https://github.com/code-kern-ai/refinery-updater/pull/28
Config: https://github.com/code-kern-ai/refinery-config/pull/20

### DIFF
--- a/business_objects/admin_message.py
+++ b/business_objects/admin_message.py
@@ -25,9 +25,10 @@ def create(text: str, level: str, archive_date: int, created_by: str):
     general.add(message, True)
 
 
-def archive(message_id: str, archived_by: str, archive_date):
+def archive(message_id: str, archived_by: str, archive_date, archived_reason):
     message = get(message_id)
     message.archived = True
     message.archived_by = archived_by
     message.archive_date = archive_date
+    message.archived_reason = archived_reason
     general.commit()
--- a/business_objects/admin_message.py
+++ b/business_objects/admin_message.py
@@ -1,35 +1,49 @@
+from typing import List
 from submodules.model.models import AdminMessage
 from . import general
 from ..session import session
+from datetime import datetime
 
 
-def get(message_id):
+def get(message_id: str) -> AdminMessage:
     return session.query(AdminMessage).filter(AdminMessage.id == message_id).first()
 
 
-def get_all():
+def get_all() -> List[AdminMessage]:
     return session.query(AdminMessage).filter().all()
 
 
-def get_all_active():
+def get_all_active() -> List[AdminMessage]:
     return session.query(AdminMessage).filter(AdminMessage.archived == False).all()
 
 
-def create(text: str, level: str, archive_date: int, created_by: str):
+def create(
+    text: str,
+    level: str,
+    archive_date: int,
+    created_by: str,
+    with_commit: bool = False,
+) -> AdminMessage:
     message = AdminMessage(
         text=text,
         level=level,
         archive_date=archive_date,
         created_by=created_by,
     )
-    general.add(message, True)
+    general.add(message, with_commit)
     return message
 
 
-def archive(message_id: str, archived_by: str, archive_date, archived_reason):
+def archive(
+    message_id: str,
+    archived_by: str,
+    archive_date: datetime,
+    archived_reason: str,
+    with_commit: bool = False,
+) -> None:
     message = get(message_id)
     message.archived = True
     message.archived_by = archived_by
     message.archive_date = archive_date
     message.archived_reason = archived_reason
-    general.commit()
+    general.flush_or_commit(with_commit)
--- a/business_objects/admin_message.py
+++ b/business_objects/admin_message.py
@@ -10,11 +10,16 @@ def get(message_id: str) -> AdminMessage:
 
 
 def get_all() -> List[AdminMessage]:
-    return session.query(AdminMessage).filter().all()
+    return session.query(AdminMessage).filter().limit(100).all()
 
 
 def get_all_active() -> List[AdminMessage]:
-    return session.query(AdminMessage).filter(AdminMessage.archived == False).all()
+    return (
+        session.query(AdminMessage)
+        .filter(AdminMessage.archived == False)
+        .limit(100)
+        .all()
+    )
 
 
 def create(
--- a/business_objects/admin_message.py
+++ b/business_objects/admin_message.py
@@ -1,0 +1,33 @@
+from submodules.model.models import AdminMessage
+from . import general
+from ..session import session
+
+
+def get(message_id):
+    return session.query(AdminMessage).filter(AdminMessage.id == message_id).first()
+
+
+def get_all():
+    return session.query(AdminMessage).filter().all()
+
+
+def get_all_active():
+    return session.query(AdminMessage).filter(AdminMessage.archived == False).all()
+
+
+def create(text: str, level: str, archive_date: int, created_by: str):
+    message = AdminMessage(
+        text=text,
+        level=level,
+        archive_date=archive_date,
+        created_by=created_by,
+    )
+    general.add(message, True)
+
+
+def archive(message_id: str, archived_by: str, archive_date):
+    message = get(message_id)
+    message.archived = True
+    message.archived_by = archived_by
+    message.archive_date = archive_date
+    general.commit()
--- a/business_objects/admin_message.py
+++ b/business_objects/admin_message.py
@@ -9,8 +9,8 @@ def get(message_id: str) -> AdminMessage:
     return session.query(AdminMessage).filter(AdminMessage.id == message_id).first()
 
 
-def get_all() -> List[AdminMessage]:
-    return session.query(AdminMessage).filter().limit(100).all()
+def get_all(limit: int = 100) -> List[AdminMessage]:
+    return session.query(AdminMessage).filter().limit(limit).all()
 
 
 def get_all_active(limit: int = 100) -> List[AdminMessage]:
--- a/business_objects/admin_message.py
+++ b/business_objects/admin_message.py
@@ -23,6 +23,7 @@ def create(text: str, level: str, archive_date: int, created_by: str):
         created_by=created_by,
     )
     general.add(message, True)
+    return message
 
 
 def archive(message_id: str, archived_by: str, archive_date, archived_reason):
--- a/business_objects/admin_message.py
+++ b/business_objects/admin_message.py
@@ -13,11 +13,11 @@ def get_all() -> List[AdminMessage]:
     return session.query(AdminMessage).filter().limit(100).all()
 
 
-def get_all_active() -> List[AdminMessage]:
+def get_all_active(limit: int = 100) -> List[AdminMessage]:
     return (
         session.query(AdminMessage)
         .filter(AdminMessage.archived == False)
-        .limit(100)
+        .limit(limit)
         .all()
     )
 
@@ -47,6 +47,8 @@ def archive(
     with_commit: bool = False,
 ) -> None:
     message = get(message_id)
+    if not message:
+        return
     message.archived = True
     message.archived_by = archived_by
     message.archive_date = archive_date
--- a/business_objects/monitor.py
+++ b/business_objects/monitor.py
@@ -12,6 +12,8 @@ def get_all_tasks(project_id: None, only_running: False):
     {get_running_embedding_tasks(project_id, only_running)}
     UNION
     {get_running_weak_supervision_tasks(project_id, only_running)}
+    UNION
+    {get_running_upload_tasks(project_id, only_running)}
     """
     return general.execute_all(query)
 
@@ -107,5 +109,24 @@ def get_running_weak_supervision_tasks(project_id: None, only_running: False):
     if only_running:
         query = query + f"""
         WHERE state = '{enums.PayloadState.CREATED.value}'
+        """
+    return query
+
+def get_running_upload_tasks(project_id: None, only_running: False):
+    query = f"""
+    SELECT id, 'upload' as task_type
+    FROM upload_task
+    """
+    if project_id and only_running: # TODO LOOKUP Weak Supervision States
+        query = query + f"""
+        WHERE project_id = '{project_id}' AND state = '{enums.UploadStates.IN_PROGRESS.value}'
+        """
+    if project_id:
+        query = query + f"""
+        WHERE project_id = '{project_id}'
+        """
+    if only_running:
+        query = query + f"""
+        WHERE state = '{enums.UploadStates.IN_PROGRESS.value}'
         """
     return query
--- a/business_objects/monitor.py
+++ b/business_objects/monitor.py
@@ -198,7 +198,7 @@ def __select_running_information_source_payloads(
     project_id: str = None, only_running: bool = False
 ) -> str:
     query = f"""
-    SELECT 'information_source' task_type, state, project_id
+    SELECT id, 'information_source' task_type, state, project_id, created_by
     FROM information_source_payload
     """
     if project_id and only_running:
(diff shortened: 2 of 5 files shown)


business_objects/admin_message.py
```diff
@@ -0,0 +1,35 @@
+from submodules.model.models import AdminMessage
+from . import general
+from ..session import session
+
+
+def get(message_id):
```
SimonDegrafKern: type

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+        """
+        )
+    if task_id:
+        query = (
+            query
+            + f"""
+        AND id = '{task_id}'
+        """
+        )
+    print(query)
```
lumburovskalina: Print should be removed

business_objects/admin_message.py
```diff
@@ -0,0 +1,54 @@
...
+
+def get_all() -> List[AdminMessage]:
+    return session.query(AdminMessage).filter().limit(100).all()
+
+
+def get_all_active() -> List[AdminMessage]:
+    return (
+        session.query(AdminMessage)
+        .filter(AdminMessage.archived == False)
+        .limit(100)
```
JWittmeyer: all_active suggests all however there is a limit included. To make it easier understandable i'd give it a parameter limit that defaults to 100 so the dev can decide if necessary to exclude a limit (plays a bit into a comment from gateway)

business_objects/admin_message.py
```diff
@@ -0,0 +1,54 @@
...
+
+def archive(
+    message_id: str,
+    archived_by: str,
+    archive_date: datetime,
+    archived_reason: str,
+    with_commit: bool = False,
+) -> None:
+    message = get(message_id)
+    message.archived = True
```
JWittmeyer, line 52: maybe check if not none before setting data to prevent errors for wrong ids

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+        UNION
+        ({__select_running_tokenization_tasks(project_id, only_running)})
+        UNION
+        ({__select_running_embedding_tasks(project_id, only_running)})
+        UNION
+        ({__select_running_weak_supervision_tasks(project_id, only_running)})
+        UNION
+        ({__select_running_upload_tasks(project_id, only_running)})
+    ) tasks
+    JOIN project p
```
JWittmeyer: INNER JOIN

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+    ON p.id = tasks.project_id
```
JWittmeyer: indended

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+from . import general
+from .. import enums
+
+
+def get_all_tasks(project_id: str = None, only_running: bool = False) -> List[Any]:
+    query = f""" 
+    SELECT tasks.*, orga.name organization_name
+    FROM (
+        ({__select_running_information_source_payloads(project_id, only_running)})
+        UNION
```
JWittmeyer: I'm assuming UNION ALL (UNION also checks if duplications exist which is unnecessary here since there is no overlap)

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+    general.commit()
+
+
+def set_information_source_payloads_to_failed(
+    project_id: str = None, payload_id: str = None, with_commit: bool = False
+) -> None:
+    query = f"""
+    UPDATE information_source_payload
+    SET state = '{enums.PayloadState.FAILED.value}'
+    WHERE state = '{enums.PayloadState.STARTED.value}'
```
JWittmeyer: this should be created since stated is just for crowd labeler (which shouldn't be cancelled since not threaded but user action based) and not e.g. lf runs

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+    """
+
+    if project_id:
+        query = (
+            query
+            + f"""
+        AND project_id = '{project_id}'
+        """
+        )
```
JWittmeyer: a bit hard to read imo. I'd go with:
```suggestion
   if project_id:
        query+=f" AND project_id = '{project_id}'"
```

If we want a linebreak \n so we don't need to write the linebreak in the triple quotes.

JWittmeyer: same for the others on the page

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+    general.flush_or_commit(with_commit)
+
+
+def set_record_tokenization_task_to_failed(
+    project_id: str = None, task_id: str = None, with_commit: bool = False
+) -> None:
+    query = f"""
+    UPDATE record_tokenization_task
+    SET state = '{enums.TokenizerTask.STATE_FAILED.value}'
+    WHERE state = '{enums.TokenizerTask.STATE_IN_PROGRESS.value}'
```
JWittmeyer: both CREATED and IN_PROGRESS since even the created ones should break if we restart.

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+    general.flush_or_commit(with_commit)
+
+
+def set_embedding_to_failed(
+    project_id: str = None, embedding_id: str = None, with_commit: bool = False
+) -> None:
+    query = f"""
+    UPDATE embedding
+    SET state = '{enums.EmbeddingState.FAILED.value}'
+    WHERE state = '{enums.EmbeddingState.ENCODING.value}' OR state = '{enums.EmbeddingState.WAITING.value}' OR state = '{enums.EmbeddingState.INITIALIZING.value}'
```
JWittmeyer: this doesn't work as expected since OR has a lower precedence than AND. We either need a brackets or (preferred) since this is on one field the IN operator

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+    general.flush_or_commit(with_commit)
+
+
+def set_weak_supervision_to_failed(
+    project_id: str = None, task_id: str = None, with_commit: bool = False
+) -> None:
+    query = f"""
+    UPDATE weak_supervision_task
+    SET state = '{enums.PayloadState.FAILED.value}'
+    WHERE state = '{enums.PayloadState.STARTED.value}'
```
JWittmeyer: payloadstate CREATED

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+    general.flush_or_commit(with_commit)
+
+
+def set_upload_task_to_failed(
+    project_id: str = None, task_id: str = None, with_commit: bool = False
+) -> None:
+    query = f"""
+    UPDATE upload_task
+    SET state = '{enums.UploadStates.ERROR.value}'
+    WHERE state = '{enums.UploadStates.IN_PROGRESS.value}' OR state = '{enums.UploadStates.PENDING.value}' OR state = '{enums.UploadStates.PREPARED.value}' OR state = '{enums.UploadStates.WAITING.value}'
```
JWittmeyer, line 128: OR issue

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+        )
+    general.execute(query)
+    general.flush_or_commit(with_commit)
+
+
+def __select_running_information_source_payloads(
+    project_id: str = None, only_running: bool = False
+) -> str:
+    query = f"""
+    SELECT id, 'information_source' task_type, state, project_id, created_by
```
JWittmeyer, line 141: should we maybe use the tablenames enum?

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+        query = (
+            query
+            + f"""
+        WHERE state = '{enums.PayloadState.CREATED.value}'
+        """
+        )
+    query = (
+        query
+        + f"""
+    LIMIT 100
```
JWittmeyer: parameter with default 100

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+    """
+    )
+    return query
```
JWittmeyer: since this will be harder and harder to extend id go with extension logic proposition (untested though :D ):
```suggestion
    where_add = ""
    if project_id:
        where_add = __extend_where(where_add, f"project_id = '{project_id}'")
    if only_running:
        where_add = __extend_where(
            where_add, f"state = '{enums.PayloadState.CREATED.value}'"
        )
    query += where_add
    query += "\nLIMIT 100"
    return query
        
        
 #function outside code block
 def __extend_where(current:str,condition:str) -> str:
    if current:
        return f"{current} AND {condition}"
    else:
        return f" WHERE {condition}"
```

JWittmeyer: similar for the others on the page

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+) -> str:
+    query = f"""
+    SELECT id, 'weak_supervision' task_type, state, project_id, created_by
+    FROM weak_supervision_task
+    """
+    if project_id and only_running:
+        query = (
+            query
+            + f"""
+        WHERE project_id = '{project_id}' AND state = '{enums.PayloadState.CREATED.value}'
```
JWittmeyer, line 364: wrong state

business_objects/monitor.py
```diff
@@ -0,0 +1,424 @@
...
+) -> str:
+    query = f"""
+    SELECT id, 'upload' task_type, state, project_id, user_id created_by
+    FROM upload_task
+    """
+    if project_id and only_running:
+        query = (
+            query
+            + f"""
+        WHERE project_id = '{project_id}' AND (state = '{enums.UploadStates.IN_PROGRESS.value}' OR state = '{enums.UploadStates.PENDING.value}' OR state = '{enums.UploadStates.PREPARED.value}' OR state = '{enums.UploadStates.WAITING.value}')
```
JWittmeyer: maybe IN operator

business_objects/user_activity.py
```diff
@@ -35,6 +41,28 @@ def get_all_user_activity() -> List[Any]:
...
+) -> models.User:
+    query = session.query(models.User)
+
+    if last_interaction_range:
+        query = session.query(models.User).filter(
+            models.User.last_interaction >= (last_interaction_range),
+        )
+
+    if order_by_interaction:
+        query = query.order_by(models.User.last_interaction.asc())
```
JWittmeyer: might be wrong but i thought this would be desc() - newest first in the list

models.py
```diff
@@ -6,12 +6,13 @@
     CascadeBehaviour,
     NotificationState,
     Tablenames,
-    Notification,
+    Notification as NotificationEnums,
```
JWittmeyer, line 9: why was this renamed?

business_objects/admin_message.py
SimonDegrafKern: done

business_objects/admin_message.py
SimonDegrafKern: done

business_objects/monitor.py
SimonDegrafKern: Done

business_objects/monitor.py
SimonDegrafKern: Done

business_objects/monitor.py
SimonDegrafKern: Done

business_objects/monitor.py
SimonDegrafKern: done

business_objects/monitor.py
SimonDegrafKern: done

SimonDegrafKern: The line breaks came from the formatter I guess IMO we don't need them

business_objects/monitor.py
SimonDegrafKern: Done

business_objects/monitor.py
SimonDegrafKern: Done

business_objects/monitor.py
SimonDegrafKern: Done

business_objects/monitor.py
SimonDegrafKern: Done

business_objects/monitor.py
SimonDegrafKern: Done

business_objects/monitor.py
SimonDegrafKern: Done

business_objects/user_activity.py
SimonDegrafKern: I think it is like this -> the first one is the newest with that implementation

models.py
SimonDegrafKern: Because before we had Notification (Enum) imported and Notification (Base) as a class defined in the same file

business_objects/monitor.py
SimonDegrafKern: Done

business_objects/monitor.py
SimonDegrafKern: I understand, this is a good solution, but maybe we wan't to tackle this if the case of needed extension occurs? Because if we ain't gonna need this I would rather keep it simple

business_objects/user_activity.py
JWittmeyer: This translates into (more or less) this db query:
image

which lists the oldest first

business_objects/monitor.py
JWittmeyer: to be honest I found the current solution irritating and it took me a bit to understand (because of multiple conditions for the same parameter & no if/elif/else logic). Also, I don't think the current solution would work since for both project_id and only_running set this would result in multiple WHERE parts. 

The proposed solution will reduce the lines of code in the whole file, increase readability and less error-prone so I'd actually say it is the simpler one (though ofc this can be a personal impression)

models.py
JWittmeyer: makes sense, can be resolved

business_objects/user_activity.py
SimonDegrafKern: Okay I am confused -> In the frontend we have it like correctly ordered, maybe there is just another ordering I did not know of, or there is something else going on I don't am aware of

business_objects/monitor.py
SimonDegrafKern: Build in your solution a litte bit adapted

business_objects/monitor.py
```diff
@@ -0,0 +1,285 @@
...
+    UPDATE {enums.Tablenames.INFORMATION_SOURCE_PAYLOAD.value}
+    SET state = '{enums.PayloadState.FAILED.value}'
+    WHERE state = '{enums.PayloadState.CREATED.value}'
+    """
+
+    if project_id:
+        query = query + f""" AND project_id = '{project_id}'"""
+
+    if payload_id:
+        query = query + f"""AND id = '{payload_id}'"""
```
JWittmeyer: i dont think this works since no space is provided so project_id and payload_id set would result in an issue.
Also, """ shouldn't be necessary for single lines.

JWittmeyer: We could also use the helper method so we don't need to worry about spaces
```suggestion
    if project_id:
        query = __extend_where_helper(query, f"project_id = '{project_id}'"

    if payload_id:
        query = __extend_where_helper(query,f"id = '{payload_id}'"
```

JWittmeyer (approved):

PR #42, opened 2023-05-16
Adds progress to attribute

SimonDegrafKern:
PRs:
- https://github.com/code-kern-ai/refinery-ui/pull/137
- https://github.com/code-kern-ai/refinery-gateway/pull/129
- https://github.com/code-kern-ai/refinery-submodule-model/pull/42
- https://github.com/code-kern-ai/refinery-ac-exec-env/pull/19

### DIFF
--- a/models.py
+++ b/models.py
@@ -397,7 +397,7 @@ class Attribute(Base):
     visibility = Column(String, default=AttributeVisibility.DO_NOT_HIDE.value)
     started_at = Column(DateTime, default=sql.func.now())
     finished_at = Column(DateTime)
-    progress = Column(Float, default=0.0)
+    progress = Column(Float)
 
     embeddings = parent_to_child_relationship(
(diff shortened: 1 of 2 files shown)


models.py
```diff
@@ -397,6 +397,7 @@ class Attribute(Base):
     visibility = Column(String, default=AttributeVisibility.DO_NOT_HIDE.value)
     started_at = Column(DateTime, default=sql.func.now())
     finished_at = Column(DateTime)
+    progress = Column(Float, default=0.0)
```
JWittmeyer: i would have assumed progress would only make sense for calculated attributes and for e.g. uploaded this would be NULL not 0.0.

SimonDegrafKern: Makes sense, will change to NULL as default

JWittmeyer (approved):

PR #44, opened 2023-06-12
Embedding providers

SimonDegrafKern:
gateway: https://github.com/code-kern-ai/refinery-gateway/pull/136
embedder: https://github.com/code-kern-ai/refinery-embedder/pull/42
gates: https://github.com/code-kern-ai/gates-runtime/pull/24
submodule model: https://github.com/code-kern-ai/refinery-submodule-model/pull/44
ui: https://github.com/code-kern-ai/refinery-ui/pull/143
admin-dashboard: https://github.com/code-kern-ai/admin-dashboard/pull/8
updater: https://github.com/code-kern-ai/refinery-updater/pull/43

### DIFF
--- a/business_objects/agreement.py
+++ b/business_objects/agreement.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 from typing import Optional
-from submodules.model import session
+from submodules.model.session import session
 from submodules.model.business_objects import general
 from submodules.model.models import Agreement
 
@@ -9,6 +9,13 @@ def get(project_id: str, item_id: str) -> Agreement:
     return (
         session.query(Agreement)
         .filter(Agreement.project_id == project_id, Agreement.id == item_id)
+        .first()
+    )
+
+def get_by_xfkey(project_id: str, xfkey: str) -> Agreement:
+    return (
+        session.query(Agreement)
+        .filter(Agreement.project_id == project_id, Agreement.xfkey == xfkey)
         .first()
     )
 
--- a/business_objects/agreement.py
+++ b/business_objects/agreement.py
@@ -12,10 +12,10 @@ def get(project_id: str, item_id: str) -> Agreement:
         .first()
     )
 
-def get_by_xfkey(project_id: str, xfkey: str) -> Agreement:
+def get_by_xfkey(project_id: str, xfkey: str, xftype: str) -> Agreement:
     return (
         session.query(Agreement)
-        .filter(Agreement.project_id == project_id, Agreement.xfkey == xfkey)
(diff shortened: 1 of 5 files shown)


business_objects/agreement.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def get(project_id: str, item_id: str) -> Agreement:
+    return (
+        session.query(Agreement)
+        .filter(Agreement.project_id == project_id, Agreement.id == item_id)
+        .first()
+    )
+
+def get_by_xfkey(project_id: str, xfkey: str) -> Agreement:
```
JWittmeyer: i think we should include a type here as well. uuid should still be unique but especially with the index on type (which is atm always embedding but might not be in the future)

business_objects/agreement.py
```diff
@@ -0,0 +1,37 @@
...
+        .first()
+    )
+
+def create(
+    project_id: str,
+    user_id: str,
+    terms_text: str,
+    terms_accepted: bool,
+    xfkey: Optional[str] = None,
+    xftype: Optional[str] = None,
```
JWittmeyer: how can these be optional? Maybe the key depending on the context but the type should always be filled

enums.py
```diff
@@ -422,3 +422,12 @@ class TaskType(Enum):
...
     WEAK_SUPERVISION = "weak_supervision"
+
+class AgreementType(Enum):
+    EMBEDDING = "emedding"
+
+class EmbeddingPlatform(Enum):
+    PYTHON = "python"
+    HUGGINGFACE = "huggingface"
+    COHERE = "cohere"
+    OPENAI = "openai"
```
JWittmeyer: why are these in lowercase? Normally we use uppercase values (with a few exceptions like table names)

business_objects/agreement.py
SimonDegrafKern: It was build like this when the order of creation was not how it is now. We first created the agreement and then the embedding in the embedding service. Now it is other way around and we now can ensure that a key is present. For the type its the same i guess

enums.py
SimonDegrafKern: I did not introduce these values but just replaced their already implemented occurrences by enums. I don't know where these values may be used otherwise for example in embedder library or other places so I did not want to change the values themselves but have a more standardized way of accessing the values

business_objects/agreement.py
SimonDegrafKern: Will include type

enums.py
JWittmeyer: Makes sense. Agreement type is new though

SimonDegrafKern: Changed embedding to EMBEDDING for agreement type

JWittmeyer (approved):

PR #45, opened 2023-06-19
Adds password support for upload task

SimonDegrafKern:
PRs:
- https://github.com/code-kern-ai/refinery-gateway/pull/137
- https://github.com/code-kern-ai/refinery-submodule-model/pull/45
- https://github.com/code-kern-ai/refinery-ui/pull/145
- https://github.com/code-kern-ai/dev-setup/pull/30


business_objects/upload_task.py
```diff
@@ -1,5 +1,5 @@
 import datetime
-from typing import Optional, Dict, Any
+from typing import ByteString, Optional, Dict, Any
```
JWittmeyer, line 2: is ByteString a typing import?

business_objects/upload_task.py
```diff
@@ -49,18 +55,39 @@ def create(
...
+    general.flush_or_commit(with_commit)
+
+
+def remove_all_keys(
+    with_commit: bool = False,
+) -> None:
+    tasks = get_all_global()
+    for task in tasks:
+        task.key = None
+    general.flush_or_commit(with_commit)
```
JWittmeyer: getting all and only using them to update one field seems like a lot of overhead, especially on a server with long living db. We should only get the ones with a key (so the list should be reduced by quite a margin) or (definitely preferred) have a small update script that clears is 
something like 
```
UPDATE upload_task
SET key = NULL
WHERE key != NULL
```

business_objects/upload_task.py
```diff
@@ -49,18 +55,39 @@ def create(
...
+
+
 def update(
     project_id: str,
     task_id: str,
     state: Optional[str] = None,
     progress: Optional[float] = None,
     file_additional_info: str = None,
     mappings: Optional[Dict[str, Any]] = None,
+    key: Optional[str] = None,
```
JWittmeyer: not sure this will work as expected if we want to update a task from key to None. 
Maybe we need something like a "NO_VALUE" as default and if it's != "NO_VALUE" we set it (even None) or the other way around and if it's == NO_VALUE we set it to None in the task.key= key part.

either way a comment would be good so the next person looking at this isn't irritated

business_objects/upload_task.py
SimonDegrafKern: yes, I searched for Byte initially but seems like ByteString is the valid option python has

business_objects/upload_task.py
SimonDegrafKern: true, will refactor it to sql

business_objects/upload_task.py
SimonDegrafKern: Yeah true, I wrote a dedicated method for removing keys but forgot to remove the parameter from this signature

business_objects/upload_task.py
JWittmeyer: makes sense, however, it's not used here

SimonDegrafKern: Will remove it

SimonDegrafKern: Added it as type for key

JWittmeyer (approved):

PR #49, opened 2023-07-20
Qdrant filter integrations 

lumburovskalina:
PRs:
- https://github.com/code-kern-ai/refinery-ui/pull/150
- https://github.com/code-kern-ai/refinery-gateway/pull/142
- https://github.com/code-kern-ai/refinery-submodule-model/pull/49
- https://github.com/code-kern-ai/refinery-neural-search/pull/31
- https://github.com/code-kern-ai/refinery-embedder/pull/52

### DIFF
--- a/business_objects/attribute.py
+++ b/business_objects/attribute.py
@@ -409,15 +409,17 @@ def get_unique_values_by_attributes(project_id: str) -> Dict[str, List[str]]:
     if not attributes:
         return {}
 
-    values_dict = {}
-    for attribute in attributes:
-        values_dict[attribute.name] = (
-            get_unique_values(project_id, attribute.name)
-            if len(get_unique_values(project_id, attribute.name)) <= 20
-            else None
-        )
+    return {
+        attribute.name: checked_unique_values(project_id, attribute.name)
+        for attribute in attributes
+    }
 
-    return values_dict
+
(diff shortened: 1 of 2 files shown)


business_objects/attribute.py
```diff
@@ -431,3 +402,30 @@ def __build_add_query(
...
+    if not attributes:
+        return {}
+
+    values_dict = {}
+    for attribute in attributes:
+        values_dict[attribute.name] = (
+            get_unique_values(project_id, attribute.name)
+            if len(get_unique_values(project_id, attribute.name)) <= 20
+            else None
+        )
```
SimonDegrafKern: I think you could reduce one call to get_unique_values(project_id, attribute.name) if you would store it before the assignment to the dictionary :) I think this could be even rewritten to a list comprehension but for this the call for the values would be outsourced into another function which would have the <= 20 logic, 

so something like this (Attention: Not working python code but python like pseudocode :D)

`return {attribute.name:checked_unique_values(project_id, attribute.name) for attribute in attributes}`

```
def checked_unique_values(project_id, attribute.name):
   value = get_unique_values(project_id, attribute.name)
   if value <= 20:
   return None
   else return value
```

JWittmeyer (approved):

PR #51, opened 2023-08-28
Embedding list enum

JWittmeyer:
- https://github.com/code-kern-ai/refinery-gateway/pull/148
- https://github.com/code-kern-ai/refinery-submodule-model/pull/51
- https://github.com/code-kern-ai/refinery-ac-exec-env/pull/27
- https://github.com/code-kern-ai/refinery-embedder/pull/58
- https://github.com/code-kern-ai/refinery-ui/pull/153
- https://github.com/code-kern-ai/refinery-neural-search/pull/36
- https://github.com/code-kern-ai/gates-runtime/pull/31

Note:
Gates was pretty much left untouched for the moment since other prs that change the structure are still in review queue & I'm unsure how we want to approach gates in the future (specify what is wanted, e.g. the new embedding_list attribute is atm fully returned etc,)

Best option to test:
`sudo bash start -g -b embedding-list` (newest dev setup version is needed -> this will start refinery & gates on the specified branch)

**otherwise:** remember to build the ac exec env & gates runtime to test :)

### DIFF
--- a/business_objects/embedding.py
+++ b/business_objects/embedding.py
@@ -200,13 +200,18 @@ def get_tensors_and_attributes_for_qdrant(
             if payload_selector != "":
                 payload_selector += ","
             if data_type != enums.DataTypes.TEXT.value:
-                payload_selector += f"'{attr}', (r.\"data\"->>'{attr}')::{data_type}"
+                payload_selector += f"'{attr}', (r.\"data\"->'{attr}')::{data_type}"
             else:
                 payload_selector += f"'{attr}', r.\"data\"->>'{attr}'"
         payload_selector = f"json_build_object({payload_selector}) payload"
+    id_column = None
+    if has_sub_key(project_id, embedding_id):
+        id_column = "et.id"
+    else:
+        id_column = "r.id"
     query = f"""
     SELECT 
-        et.id::TEXT || CASE WHEN sub_key IS NULL THEN '' ELSE '@' || sub_key::TEXT END id, 
+        {id_column}::TEXT, 
         et."data", 
         {payload_selector}
     FROM embedding_tensor et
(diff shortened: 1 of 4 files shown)


business_objects/embedding.py
```diff
@@ -200,12 +201,20 @@ def get_tensors_and_attributes_for_qdrant(
             if payload_selector != "":
                 payload_selector += ","
             if data_type != enums.DataTypes.TEXT.value:
-                payload_selector += f"'{attr}', (r.\"data\"->>'{attr}')::{data_type}"
+                payload_selector += f"'{attr}', (r.\"data\"->'{attr}')::{data_type}"
```
jhoetter, line 204: minor hint: the difference between -> and ->> are not clear to me. Again, just for the dumber programmers (me), the comment why the change is relevant (or even _that_ there is a change) would have been cool for me :-)

business_objects/embedding.py
```diff
@@ -215,6 +224,39 @@ def get_tensors_and_attributes_for_qdrant(
...
+) -> int:
+    query = f"""
+    SELECT CEIL(AVG(max_key))::INTEGER
+    FROM (
+        SELECT record_id, MAX(sub_key + 1) max_key
+        FROM embedding_tensor et
+        WHERE project_id = '{project_id}' AND embedding_id = '{embedding_id}'
+        GROUP BY record_id )x """
+    value = general.execute_first(query)
+    if value is None or value[0] is None:
```
jhoetter, line 255: not sure: can it be that value is `[]`? or is that never the case? if it can be, a check for list length could be reasonable. If it can't be, just resolve :-)

business_objects/embedding.py
```diff
@@ -355,16 +419,30 @@ def create_tensors(
...
-            project_id=project_id,
-            record_id=record_id,
-            embedding_id=embedding_id,
-            data=tensor,
-        )
-        for record_id, tensor in zip(record_ids, tensors)
-    ]
-    general.add_all(tensors, with_commit)
+    to_add = None
+    if len(record_ids) > 0 and "@" in record_ids[0]:
```
jhoetter, line 423: what is the meaning of the "@" :D (sorry if I'm adding all the comments regarding readability, but to me it isn't clear what the context is directly)

jhoetter: is it meant as something to split the text?

jhoetter (approved):
only few comments regarding documentation, looks good to me and works

business_objects/embedding.py
JWittmeyer: in short previous version was faulty. -> returns a json object ->> directly casts to string. Since we cast after that anyway the second > was unnecessary from the initial introduction

JWittmeyer: also i'm not sure if we should start explaining operators in comments because there will be no end to it

business_objects/embedding.py
JWittmeyer: value is a scalar through execute_first. the [0] ist access of the first entry first columvalue so -> can't be an empty list

business_objects/embedding.py
JWittmeyer: added comment & yes

PR #67, opened 2023-12-05
Unprototyping Cognition

FelixKirschKern:
Link to main PR: https://github.com/code-kern-ai/cognition-gateway/pull/23.
Branched from `refactor-response` ~~as I expect that it's merged into `dev` soon~~, but will be merged to dev (merge then also includes the changes from `refactor-response`)


util.py
```diff
@@ -116,6 +118,10 @@ def to_frontend_obj(value: Union[List, Dict]):
         return {__to_camel_case(k): to_frontend_obj(v) for k, v in value.items()}
     elif is_list_like(value):
         return [to_frontend_obj(x) for x in value]
+    elif isinstance(
+        value, BaseModel
+    ):  # TODO: fix to get it working now but better be placed in client_response.py
+        return to_json(value)
```
FelixKirschKern: This was added because I ran into errors when I have a pydantic object that is nested in a dictionary.

FelixKirschKern: A related problem is that currently keys are changed to camel case that should not be changed, for example embedding names.

JWittmeyer: not sure about related probelm since the logic shouldn't change values only keys. Might also misunderstand it :D

JWittmeyer (approved):

PR #73, opened 2024-01-26
Adds sql injection prevention

JWittmeyer:
- https://github.com/code-kern-ai/refinery-submodule-model/pull/73
- https://github.com/code-kern-ai/gates-gateway/pull/45
- https://github.com/code-kern-ai/cognition-gateway/pull/35
- https://github.com/code-kern-ai/refinery-gateway/pull/189

Includes sql injection prevention for all model files

For the session [handling PR](https://github.com/code-kern-ai/refinery-gateway/pull/188) the submodules where already exchanged to test with an attack vector.

______________________________
possible extension (not in this PR)
- optimize prevent method to get any input type to work easier with list, dicts etc.
- maybe add prevention earlier (e.g. in middleware?)
- maybe use a different prevention method (e.g. prepared statements)
- generated selects from db values (e.g. attribute names) might cause issues in edgecases and should also be sanitized.

### DIFF
--- a/business_objects/export.py
+++ b/business_objects/export.py
@@ -5,12 +5,18 @@ from .. import enums
 from ..business_objects.payload import get_base_query_valid_labels_manual
 from ..models import Attribute
 from . import user_session
+from ..util import prevent_sql_injection
 
 OUTSIDE_CONSTANT = "OUTSIDE"
+
 
 def build_full_record_sql_export(
     project_id: str, attributes: List[Attribute], user_session_id: str
 ) -> str:
+    project_id = prevent_sql_injection(project_id, isinstance(project_id, str))
+    user_session_id = prevent_sql_injection(
+        user_session_id, isinstance(user_session_id, str)
+    )
     select_part: str = ""
     for att in attributes:
         if select_part != "":
@@ -29,7 +35,7 @@ def build_full_record_sql_export(
 
     if select_add:
         select_part += ",\n" + select_add
-    return get_final_sql(
+    return __get_final_sql(
         project_id,
         select_part,
         classification_part,
@@ -72,7 +78,7 @@ def __build_user_session_part(project_id: str, user_session_id: str) -> Tuple[st
 
 
 def __build_classification_part(project_id: str) -> Tuple[str, str]:
-    classification_case = get_case_classification(project_id)
+    classification_case = __get_case_classification(project_id)
     if not classification_case:
         return "", ""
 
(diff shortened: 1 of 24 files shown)


business_objects/tokenization.py
```diff
@@ -79,9 +81,14 @@ def get_doc_bin_progress(project_id: str) -> str:
 def get_doc_bin_table_to_json(
     project_id: str, missing_columns: str, record_ids: List[str] = None
 ) -> Any:
+    project_id = prevent_sql_injection(project_id, isinstance(project_id, str))
```
LennartSchmidtKern, line 84: If we follow an approach of sanitizing input possible given by users, we should here maybe also sanitize the missing_columns because the value can be set by users through attribute naming

business_objects/record.py
```diff
@@ -194,6 +207,10 @@ def update_bytes_of_record_tokenized(
 
 
 def update_columns_of_tokenized_records(rt_ids: str, attribute_name: str) -> None:
+    # rt_ids = prevent_sql_injection(rt_ids, isinstance(rt_ids, str)) # excluded since prepared with ()
```
LennartSchmidtKern: How does the preperation with "()" avoid something like "(" + "'1' OR '1'='1') AND ('2'='2'" + ")"

JWittmeyer: maybe misleading comment. this isn't relevant as the rt_ids string is prepared beforehand. (in refinery-tokenizer)
image

the base assumtion on which the prevent method works is that it's a value not a full sql string or similar. So maybe as comment
`excluded since already prepared beforehand`

business_objects/tokenization.py
JWittmeyer: collected from database & prepared beforehand. Added comment as indicator

business_objects/labeling_access_link.py
```diff
@@ -72,6 +73,7 @@ def get_all_by_type_and_external_id(
 
 
 def __get_add_ids_data_slice(project_id: str, slice_id: str) -> List[str]:
```
LennartSchmidtKern, line 75: I think slice_id must also be sanitized as it can come from a query mutate

business_objects/export.py
```diff
@@ -29,7 +35,7 @@ def build_full_record_sql_export(
 
     if select_add:
         select_part += ",\n" + select_add
-    return get_final_sql(
+    return __get_final_sql(
```
LennartSchmidtKern, line 38: Such big sql assembly functions with lot of private functions seem to be hard to control. On first view, it seems to be ok by only sanitzing the project_id on the outer level here. I think we have also to consider the intermediate processing of user controllable values. For example: 1. query could select some labeling task and second query uses the labelink task name as variable in f string. Then we would still need to sanitze the column name I would assume.

business_objects/labeling_access_link.py
JWittmeyer: 100% agree, changed

business_objects/export.py
JWittmeyer: understandable, ill put it under further extensions in the PR.
I think the most common case for that is the attributes name as it is the only one that is used in queries directly (not select but as access). (Used e.g. in the database as the json value selection in postgres needs the name)

LennartSchmidtKern (approved):

PR #75, opened 2024-02-09
Authorization via Teams

FelixKirschKern:
Link to Main PR: https://github.com/code-kern-ai/refinery-gateway/pull/182


business_objects/team_member.py
```diff
@@ -0,0 +1,74 @@
...
+    general.add(team_member, with_commit)
+    return team_member
+
+
+def delete_by_team_and_user_id(
+    team_id: str, user_id: str, with_commit: bool = False
+) -> None:
+    team_member = get_by_team_and_user(team_id, user_id)
+    if team_member:
+        general.delete(team_member, with_commit)
```
lumburovskalina, line 65: not sure but i think `general.flush_or_commit(with_commit)` is missing here

FelixKirschKern: `flush_or_commit` is handled inside of `general.delete`

lumburovskalina (approved):

anmarhindi (approved):

business_objects/team_member.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def delete_by_user_id(user_id: str, with_commit: bool = False) -> None:
+    team_memberships = (
+        session.query(TeamMember).filter(TeamMember.user_id == user_id).all()
+    )
+    for membership in team_memberships:
+        general.delete(membership, with_commit=False)
+    general.flush_or_commit(with_commit)
```
JWittmeyer, line 74: not needed

business_objects/team_member.py
```diff
@@ -0,0 +1,74 @@
...
+    if team_member:
+        general.delete(team_member, with_commit)
+
+
+def delete_by_user_id(user_id: str, with_commit: bool = False) -> None:
+    team_memberships = (
+        session.query(TeamMember).filter(TeamMember.user_id == user_id).all()
+    )
+    for membership in team_memberships:
+        general.delete(membership, with_commit=False)
```
JWittmeyer, line 73: should be 
```suggestion
        general.delete(membership, with_commit=with_commit)
```

business_objects/team_member.py
```diff
@@ -0,0 +1,74 @@
...
+    already_exist = get_by_team_and_user(team_id=team_id, user_id=user_id)
+    if already_exist:
+        return already_exist
+
+    team_item = team.get(team_id)
+    user_item = user.get(user_id)
+    if not team_item or not user_item:
+        return None
+    if team_item.organization_id != user_item.organization_id:
+        return None
```
JWittmeyer: why not raise? this is something that shouln't happen (assuming only users that are in the org are selectable)

business_objects/team_resource.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+    general.add(team_resource, with_commit)
+    return team_resource
+
+
+def __get_resource(resource_id: str, resource_type: enums.ResourceType):
+    if resource_type == enums.ResourceType.COGNITION_PROJECT:
+        return cognition_project.get(resource_id)
+    else:
+        return None
```
JWittmeyer: why not raise? this shouldn't be happening => raise is usually a good hint for the dev :)

cognition_objects/project.py
```diff
@@ -14,19 +14,51 @@ def get(project_id: str) -> CognitionProject:
...
-        .order_by(CognitionProject.created_at.asc())
-        .all()
+        .join(TeamResource, TeamResource.resource_id == CognitionProject.id)
+        .join(TeamMember, TeamMember.team_id == TeamResource.team_id)
+        .filter(TeamMember.user_id == user_id)
+        .filter(
+            TeamResource.resource_type == enums.ResourceType.COGNITION_PROJECT.value
+        )
+        .filter(CognitionProject.id == project_id)
+        .first()
```
JWittmeyer: maybe add comment to ensure it's clear that an engineer calling this method wont get anything

cognition_objects/project.py
```diff
@@ -14,19 +14,51 @@ def get(project_id: str) -> CognitionProject:
...
+        .join(TeamMember, TeamMember.team_id == TeamResource.team_id)
+        .filter(TeamMember.user_id == user_id)
+        .filter(
+            TeamResource.resource_type == enums.ResourceType.COGNITION_PROJECT.value
+        )
+        .filter(CognitionProject.organization_id == org_id)
+        .order_by(CognitionProject.created_at.asc())
+        .all()
+    )
+
```
JWittmeyer: i feel like this could cause issues. (Maybe some misconception i have)

i read this as "get me all projects user x has access to". If i understand the logic right this only works if add all engineers  are also part of the team tables.

So maybe a differnt function name or comment to clarify that it's not "get me all projects user x has access to"

enums.py
```diff
@@ -146,6 +149,10 @@ def snake_case_to_camel_case(self):
         )
 
 
+class ResourceType(Enum):
```
JWittmeyer: maybe TeamResourceType

business_objects/team_member.py
FelixKirschKern: changed

business_objects/team_resource.py
FelixKirschKern: changed

enums.py
FelixKirschKern: renamed

cognition_objects/project.py
FelixKirschKern: I see the problem. I decided to move the check for the user role inside the submodule model. So, now `get_all_by_user` and `get_by_user` work for all user roles. When these methods are used, it is checked if the given user has access, considering his role.
So, if the submodule is used somewhere we don't need to write the code for different behaviour for each role again. I feel like this is clearer and safer now. Do you agree?

cognition_objects/project.py
FelixKirschKern: see reply to comment on `get_all_by_user`

cognition_objects/project.py
JWittmeyer: i like it :)

PR #80, opened 2024-02-26
adds project filter

JWittmeyer:


FelixKirschKern (approved):
Fixes the issue.

But I think we should not let the organization_id field empty at anytime, we could distinguish between org and project level by checking if the project_id is set (or null). This reduces complexity, and could prevent issues in the future. We would need to update existing data, so maybe nothing for now.

PR #89, opened 2024-04-11
Custom Markdown Renderer

LennartSchmidtKern:
Related PR
Main PR:
- https://github.com/code-kern-ai/cognition-ui/pull/41

### DIFF
--- a/business_objects/attribute.py
+++ b/business_objects/attribute.py
@@ -13,6 +13,13 @@ from submodules.model import enums
 from ..util import prevent_sql_injection
 
 
+DEFAULT_ATTRIBUTE_STATES_USEABLE = [
+    AttributeState.UPLOADED.value,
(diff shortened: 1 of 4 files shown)


business_objects/attribute.py
```diff
@@ -116,6 +116,24 @@ def get_text_attributes(
     return {att.name: str(att.id) for att in text_attributes}
 
 
+def get_category_attributes(
+    project_id: str,
+    state_filter: List[str] = [
+        AttributeState.UPLOADED.value,
+        AttributeState.USABLE.value,
+        AttributeState.AUTOMATICALLY_CREATED.value,
+    ],
```
JWittmeyer: antipatter in python as the default variable is only filled once so changes to it will result in unpredictable behaviour

JWittmeyer (approved):

PR #91, opened 2024-04-22
tf-idf embeddings for search

JWittmeyer:
main PR: https://github.com/code-kern-ai/refinery-embedder/pull/89

### DIFF
--- a/business_objects/record.py
+++ b/business_objects/record.py
@@ -825,7 +825,10 @@ def __get_order_by(project_id: str, first_x: int = 3) -> str:
         if order != "":
             order += ", "
         tmp = f"data->>'{x.name}'"
-        if x.data_type == "INTEGER":
+
+        r_id = attribute.get_running_id_name(project_id)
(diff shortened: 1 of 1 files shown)


business_objects/record.py
```diff
@@ -825,7 +825,10 @@ def __get_order_by(project_id: str, first_x: int = 3) -> str:
         if order != "":
             order += ", "
         tmp = f"data->>'{x.name}'"
-        if x.data_type == "INTEGER":
+
+        r_id = attribute.get_running_id_name(project_id)
```
LennartSchmidtKern, line 829: What is the purpose of ordering the records for embedding creation, what would change if we just remove it?

JWittmeyer: It was initially created to have some comparison between systems as there are multiple factors changing the "final" values. One of them is the order in wich the records come in. This shouldn't change the result in a meaningfull way only make it more compareable. At least that was the idea when it was implemented.

LennartSchmidtKern (approved):

PR #103, opened 2024-05-26
Pagination users

lumburovskalina:
Main PR: https://github.com/code-kern-ai/admin-dashboard/pull/26

### DIFF
--- a/business_objects/user_activity.py
+++ b/business_objects/user_activity.py
@@ -103,3 +103,9 @@ def __remove_old_user_activity_entries(
     WHERE ua.id = helper.id; """
     general.execute(query)
     general.flush_or_commit(with_commit)
+
+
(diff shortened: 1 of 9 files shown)


business_objects/user_activity.py
```diff
@@ -103,3 +103,9 @@ def __remove_old_user_activity_entries(
     WHERE ua.id = helper.id; """
     general.execute(query)
     general.flush_or_commit(with_commit)
+
+
+def delete_user_activity(user_id: str, with_commit: bool = False) -> None:
+    query = f"DELETE FROM user_activity WHERE created_by = '{user_id}'"
```
LennartSchmidtKern, line 111: Even if this is a admin endpoint, we should sanitize this SQL input I would suggest

LennartSchmidtKern (approved):

PR #105, opened 2024-05-31
ETL different providers

lumburovskalina:
Main PR: https://github.com/code-kern-ai/cognition-ui/pull/57

### DIFF
--- a/cognition_objects/environment_variable.py
+++ b/cognition_objects/environment_variable.py
@@ -63,7 +63,12 @@ def get_by_name_and_org_id(
 
 
 def get_by_md_file_id(md_file_id: str) -> CognitionEnvironmentVariable:
-    env_var_id = cast(CognitionMarkdownDataset.llm_config.op("->>")("envVarId"), UUID)
+    env_var_id = cast(
+        CognitionMarkdownDataset.llm_config.op("->")("transformation").op("->>")(
+            "envVarId"
+        ),
+        UUID,
+    )
     return (
         session.query(CognitionEnvironmentVariable)
         .join(
@@ -83,7 +88,12 @@ def get_by_md_file_id(md_file_id: str) -> CognitionEnvironmentVariable:
 
 def get_dataset_env_var_value(dataset_id: str, org_id) -> CognitionEnvironmentVariable:
 
-    env_var_id = cast(CognitionMarkdownDataset.llm_config.op("->>")("envVarId"), UUID)
+    env_var_id = cast(
+        CognitionMarkdownDataset.llm_config.op("->")("transformation").op("->>")(
+            "envVarId"
+        ),
+        UUID,
+    )
     return (
         session.query(CognitionEnvironmentVariable)
         .join(
@@ -103,7 +113,10 @@ def get_dataset_azure_models_env_var_value(
 ) -> CognitionEnvironmentVariable:
 
     env_var_id = cast(
-        CognitionMarkdownDataset.llm_config.op("->>")("azureDiEnvVarId"), UUID
+        CognitionMarkdownDataset.llm_config.op("->")("extraction").op("->>")(
+            "azureDiEnvVarId"
+        ),
+        UUID,
     )
     return (
         session.query(CognitionEnvironmentVariable)
@@ -131,7 +144,10 @@ def get_all_by_project_id(project_id: str) -> List[CognitionEnvironmentVariable]
 
 def get_cognition_project_env_var_value(cognition_project_id: str) -> str:
 
-    env_var_id = cast(CognitionProject.llm_config.op("->>")("envVarId"), UUID)
+    env_var_id = cast(
+        CognitionProject.llm_config.op("->")("transformation").op("->>")("envVarId"),
+        UUID,
+    )
     v = (
         session.query(CognitionEnvironmentVariable.value)
         .join(CognitionProject, env_var_id == CognitionEnvironmentVariable.id)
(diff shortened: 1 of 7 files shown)


cognition_objects/environment_variable.py
```diff
@@ -46,13 +48,27 @@ def get_by_name(
...
+def get_by_name_and_org_id(
+    org_id: str,
+    name: str,
+) -> CognitionEnvironmentVariable:
+
+    return (
+        session.query(CognitionEnvironmentVariable)
+        .filter(CognitionEnvironmentVariable.organization_id == org_id)
+        .filter(CognitionEnvironmentVariable.project_id == None)
+        .filter(CognitionEnvironmentVariable.name == name)
```
JWittmeyer: usually we use filter with multiple comma sperated filters for AND connected ones we dont need to extend afterwards

cognition_objects/environment_variable.py
```diff
@@ -76,12 +130,25 @@ def get_all_by_project_id(project_id: str) -> List[CognitionEnvironmentVariable]
...
-        .join(
-            CognitionProject,
-            CognitionProject.open_ai_env_var_id == CognitionEnvironmentVariable.id,
+        .join(CognitionProject, env_var_id == CognitionEnvironmentVariable.id)
+        .filter(
+            CognitionProject.id == cognition_project_id,
         )
+        .first()
+    )
+    if v:
```
JWittmeyer: i think v can also exists if the result is NULL so i usually check `v and v[0]` to prevent str(None) = 'None'

cognition_objects/environment_variable.py
```diff
@@ -46,13 +48,36 @@ def get_by_name(
...
+        session.query(CognitionEnvironmentVariable)
+        .filter(
+            and_(
+                CognitionEnvironmentVariable.organization_id == org_id,
+                CognitionEnvironmentVariable.project_id == None,
+                CognitionEnvironmentVariable.name == name,
+            )
+        )
+        .first()
+    )
```
JWittmeyer: why do we write the default and_ in this filter? Usually we use it directly:
```suggestion
    return (
        session.query(CognitionEnvironmentVariable)
        .filter(            
                CognitionEnvironmentVariable.organization_id == org_id,
                CognitionEnvironmentVariable.project_id == None,
                CognitionEnvironmentVariable.name == name,            
        )
        .first()
    )
```

cognition_objects/markdown_file.py
```diff
@@ -202,6 +202,7 @@ def create_md_llm_log(
     created_at: Optional[datetime] = None,
     finished_at: Optional[datetime] = None,
     with_commit: bool = True,
+    delayed_commit: bool = False,
 ) -> None:
```
JWittmeyer: return type isn't correct anymore for delayed commit (which is something i haven'T seen yet)

What exactly should it do?

Usually we always return the object, with_commit only changes if the values are just flushed (prefilled defaults filled, not yet in db) or commited. 

In this case the create_md_llm_log appears to be still in a prototyped state created without our common dev rules in mind. I think we should change it to always return the object like all our other create methods do & remove the delayed commit parameter.

cognition_objects/project.py
```diff
@@ -331,6 +332,23 @@ def update(
...
+    if llm_config is not None:
+        new_values = project.llm_config
+        if new_values is None:
+            new_values = {}
+        for key in llm_config:
+            if isinstance(llm_config[key], dict):
+                if key not in new_values:
+                    new_values[key] = {}
+                for sub_key in llm_config[key]:
+                    new_values[key][sub_key] = llm_config[key][sub_key]
```
JWittmeyer, line 346: this now allows a max depth of two. I think for now this is fine. However with this approach we need to remember this / have to find it if problematic. 

I usually prefer options we don't need to remember since a lot of things are forgotten over time. Something like the `transferNestedDict` from the frontend could do the trick on the backend as well so we have a util method we can always use in these instances. (Not necessary for this feature but the next time a nested dict needs to be transferred i think we should invest the time)

cognition_objects/markdown_file.py
anmarhindi: The delayed_commit allows the CognitionMarkdownLLMLogs values to be returned (md_llm_logs) and not committed immediately due to the parallelisation added for it. This gives us a speed improvement but must be handled in a way that not too many isolated commits are done, but all in sequence or a batch in the end, otherwise we encounter issues.

For us to use the parallelised `process_text_block` we must delay the commits.

JWittmeyer: that's exactly what with_commit=False is used for/does. Unprototyped function without delayed_commit please.

anmarhindi: Ok just to confirm, I am adjusting the create_md_llm_log function only here? I initially thought you meant to replace it in the whole stack

JWittmeyer: not exactly, the change should also include our common create method logic which follows the pattern:

``` python
# class object creation
conversation: CognitionConversation = CognitionConversation(
        project_id=project_id,
        created_by=user_id,
        created_at=created_at,
        has_tmp_files=has_tmp_files,
        scope_dict={},
)
# commit or flush (add calls flush_or_commit internally)  (flushed = default like id are already set correctly, not yet in db)
general.add(conversation, with_commit)
# return object
return conversation
```

so the change is an unprototyped version of the create method for markdown files which then can be used like intendet.

cognition_objects/project.py
anmarhindi: Comment added for extension if needed in the future

cognition_objects/markdown_file.py
anmarhindi: Great. I found a clean working solution that solves the issues and retains the speed gained.

JWittmeyer (approved):

PR #108, opened 2024-07-02
Adds customer button table

JWittmeyer:
main PR: https://github.com/code-kern-ai/refinery-gateway/pull/232

### DIFF
--- a/global_objects/customer_button.py
+++ b/global_objects/customer_button.py
@@ -18,21 +18,32 @@ def get(id: str) -> CustomerButton:
 
 
 def get_all(filter_visible: Optional[bool] = None) -> List[CustomerButton]:
-    if not filter_visible:
-        return session.query(CustomerButton).all()
 
-    return (
-        session.query(CustomerButton)
-        .filter(CustomerButton.visible == filter_visible)
-        .all()
-    )
+    where_add = ""
+
+    if filter_visible is not None:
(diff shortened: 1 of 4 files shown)


global_objects/customer_button.py
```diff
@@ -0,0 +1,109 @@
...
+    visible: Optional[bool] = None,
+    with_commit: bool = True,
+) -> CustomerButton:
+    button = get(id)
+    if not button or str(org_id) != str(button.organization_id):
+        raise ValueError("Button not found or not part of the organization")
+    if type:
+        button.type = type.value
+    if location:
+        button
```
LennartSchmidtKern: probably `button.location = location`

global_objects/customer_button.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+        .first()
+    )
+
+
+def get_all(filter_visible: Optional[bool] = None) -> List[CustomerButton]:
+
+    where_add = ""
+
+    if filter_visible is not None:
```
LennartSchmidtKern, line 25: maybe we should sanitize the filter visible, even if with the current implementation boolean is enforeced on the request level

global_objects/customer_button.py
JWittmeyer: `location.value` but yes :)

LennartSchmidtKern (approved):

PR #109, opened 2024-07-04
Macros with Folder

LennartSchmidtKern:
Main PR:
- https://github.com/code-kern-ai/cognition-gateway/pull/102

### DIFF
--- a/cognition_objects/macro.py
+++ b/cognition_objects/macro.py
@@ -41,7 +41,7 @@ def get_with_nodes_and_edges(macro_id: str) -> Dict[str, Any]:
     query = f"""
     SELECT row_to_json(x)
     FROM (
-        SELECT M.*, me.edges, mn.nodes
+        SELECT M.*,  COALESCE(me.edges, '{{}}'::json[]) AS edges, COALESCE(mn.nodes, '{{}}'::json[]) AS nodes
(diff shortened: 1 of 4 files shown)


models.py
```diff
@@ -1132,6 +1132,7 @@ class CognitionProject(Base):
 
     allow_file_upload = Column(Boolean, default=False)
     max_file_size_mb = Column(Float, default=3.0)
+    max_folder_size_mb = Column(Float, default=20.0)
```
lumburovskalina, line 1135: i think this is not float anymore since we have a dropdown with options

LennartSchmidtKern: It is not relevant for the database, it is still a float there. I would keep this data type here to give flexibility in the future. We now changed the input to dropdown predefined values, but still could go to mor arbitrary input.

cognition_objects/macro.py
```diff
@@ -458,20 +458,19 @@ def get_all_executions_by_group_id(
     )
 
 
-def get_macro_execution_overview_for_document_message_queue(
+def get_macro_execution_overview_for_document_or_folder_message_queue(
```
lumburovskalina: see comment for renaming on cognition

lumburovskalina (approved):

PR #117, opened 2024-08-08
Neural Search step type

JWittmeyer:
main PR: https://github.com/code-kern-ai/cognition-gateway/pull/122

### DIFF
--- a/business_objects/embedding.py
+++ b/business_objects/embedding.py
@@ -149,6 +149,19 @@ def get_waiting_embeddings(project_id: str) -> List[Embedding]:
     )
 
 
+def get_finished_embeddings_dropdown_list(project_id: str) -> List[Dict[str, str]]:
+    project_id = prevent_sql_injection(project_id, isinstance(project_id, str))
+    query = f"""
+    SELECT array_agg(jsonb_build_object('value', e.id,'name',e.NAME))
+    FROM embedding e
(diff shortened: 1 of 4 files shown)


business_objects/embedding.py
```diff
@@ -327,6 +340,74 @@ def get_match_record_ids_to_qdrant_ids(
     return [r[0] for r in general.execute_all(query)]
 
 
+def get_match_record_ids_to_qdrant_ids_with_max_score(
+    project_id: str, embedding_id: str, qdrant_results: List[Any], limit: int
+) -> List[Any]:
```
lumburovskalina, line 345: not sure if it makes sense but usually we have a default value for limit?

business_objects/embedding.py
```diff
@@ -327,6 +340,74 @@ def get_match_record_ids_to_qdrant_ids(
...
+    GROUP BY et.record_id
+    ORDER BY 2 ASC
+    LIMIT {limit}
+    """
+
+    return [{"id": r[0], "score": r[1]} for r in general.execute_all(query)]
+
+
+def __generate_with_table_union_query(qdrant_results: List[Any]) -> str:
+
```
lumburovskalina, line 393: prevent_sql_injection for qdrant_results?

business_objects/embedding.py
JWittmeyer: not in this case as it's a required parameter calculated by the embedding or neural search service

business_objects/embedding.py
JWittmeyer: not needed as the values aren't from outside & not string based but a return object class from qdrant itself.

lumburovskalina (approved):

PR #120, opened 2024-08-27
Smart Planner Unprototype

LennartSchmidtKern:
Main PR:
- https://github.com/code-kern-ai/cognition-gateway/pull/133


models.py
```diff
@@ -1119,7 +1119,7 @@ class CognitionProject(Base):
     name = Column(String)
     description = Column(String)
     color = Column(String)
-    operator_routing_source_code = Column(String)
+    operator_routing_config = Column(JSON)
```
JWittmeyer, line 1122: renaming a column often results in loss of data in default alembic as it usually works with drop & add (also in refinery-gateway note)

JWittmeyer (approved):

PR #130, opened 2024-10-08
Cache Files for ETL Processing

LennartSchmidtKern:
Main PR:
- https://github.com/code-kern-ai/cognition-ui/pull/98


business_objects/monitor.py
```diff
@@ -7,8 +7,15 @@
 from submodules.model.cognition_objects import (
     macro as macro_db_bo,
     markdown_file as markdown_file_db_bo,
+    file_extraction as file_extraction_db_bo,
+    file_transformation as file_transformation_db_bo,
 )
 
+FILE_CACHING_IN_PROGRESS_STATES = [
```
JWittmeyer, line 14: maybe in the future we should use sets instread of lists for IN checks. At least from a theoretically standpoint a set would be faster for `in` checks. In practice with two entries it shouldn't even be noticable => nothing to do atm

cognition_objects/file_extraction.py
```diff
@@ -0,0 +1,113 @@
...
+            FileExtraction.file_reference_id == file_reference_id,
+            FileExtraction.extraction_key == extraction_key,
+            (
+                FileExtraction.state == enums.FileCachingState.COMPLETED.value
+                if only_completed
+                else True
+            ),
+        )
+        .first()
+    )
```
JWittmeyer: usually we write this as a query extension since it's easier to read & easier to extend if needed

```suggestion

    query = session.query(FileExtraction)
        .filter(
            FileExtraction.organization_id == org_id,
            FileExtraction.file_reference_id == file_reference_id,
            FileExtraction.extraction_key == extraction_key,
        )
    
    if only_completed:
        query = query.filter( FileExtraction.state == enums.FileCachingState.COMPLETED.value)
    
    return query.first()
 
```

cognition_objects/file_extraction.py
```diff
@@ -0,0 +1,113 @@
...
+    general.add(file_extraction, with_commit)
+
+    return file_extraction
+
+
+def update(
+    org_id: str,
+    file_extraction_id: str,
+    minio_path: str = None,
+    state: str = None,
```
JWittmeyer: = None have typed `Optional[str] = None`

cognition_objects/file_transformation.py
```diff
@@ -0,0 +1,116 @@
...
+        session.query(FileTransformation)
+        .filter(
+            FileTransformation.organization_id == org_id,
+            FileTransformation.file_extraction_id == file_extraction_id,
+            FileTransformation.transformation_key == transformation_key,
+            (
+                FileTransformation.state == enums.FileCachingState.COMPLETED.value
+                if only_completed
+                else True
+            ),
```
JWittmeyer: see above

cognition_objects/file_transformation.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+        .first()
+    )
+
+
+def get_by_id(org_id, file_transformation_id: str) -> FileTransformation:
```
JWittmeyer: typing

cognition_objects/file_transformation.py
```diff
@@ -0,0 +1,116 @@
...
+    general.add(file_transformation, with_commit)
+
+    return file_transformation
+
+
+def update(
+    org_id: str,
+    file_transformation_id: str,
+    minio_path: str = None,
+    state: str = None,
```
JWittmeyer: see above

JWittmeyer (approved):

PR #141, opened 2024-11-20
Pipeline Versions

JWittmeyer:
main PR: https://github.com/code-kern-ai/cognition-ui/pull/118


models.py
```diff
@@ -1253,6 +1254,14 @@ class CognitionMessage(Base):
 
     scope_dict_diff_previous_conversation = Column(JSON)
     scope_dict_diff_new = Column(JSON)
+    version_id = Column(
+        # pipeline version with which the version was created, can be null if the version isn't available anymore (e.g. deleted)
```
LennartSchmidtKern: version= message in the second part here?

cognition_objects/pipeline_version.py
```diff
@@ -0,0 +1,99 @@
...
+MAX_AUTO_SAVE_VERSIONS = 10
+
+
+def get_current_version(project_id: str) -> Optional[CognitionPipelineVersion]:
+    return (
+        session.query(CognitionPipelineVersion)
+        .filter(
+            CognitionPipelineVersion.project_id == project_id,
+        )
+        .order_by(CognitionPipelineVersion.created_at.desc())
```
LennartSchmidtKern, line 16: Is the assumption current -> created_at desc always true? If we set our current version to some older version e.g.

models.py
JWittmeyer: no the actual pipeline version is meant so if it's deleted this will cascade to null. 

Maybe I'm misunderstanding the question 🤔

cognition_objects/pipeline_version.py
JWittmeyer: yes it's true (also a reason why the activate creates a new version :)

models.py
LennartSchmidtKern: pipeline version with which the **version** was created = pipeline version with which the **message** was created.

Or what is the version, the  version of the message was created with?

enums.py
```diff
@@ -791,3 +794,8 @@ class ChangeAction(Enum):
     LIST_VALUE_CHANGED = "LIST_VALUE_CHANGED"
     ITEM_REMOVED = "ITEM_REMOVED"
     ITEM_ADDED = "ITEM_ADDED"
+
+
+class PipelineVersionType(Enum):
+    AUTO_SAVE = "AUTO_SAVE"  # any save operation in relevant but only 10 per project
+    PERSISTED = "NAMED_VERSION"  # any AUTO_SAVE that is considered worth keeping
```
LennartSchmidtKern: usually our enum values are the same as the name, why not "PERSISTED"?

models.py
JWittmeyer: got it :D changed

enums.py
JWittmeyer: changed to named version

LennartSchmidtKern (approved):

PR #143, opened 2024-11-25
Custom cleanup conversations and files by org 

LennartSchmidtKern:
Main PR:
- https://github.com/code-kern-ai/refinery-gateway/pull/275


cognition_objects/conversation.py
```diff
@@ -25,6 +25,20 @@ def get(project_id: str, conversation_id: str) -> CognitionConversation:
...
+    query = """
+    SELECT
+        cc.*,
+        cp.organization_id
+    FROM
+        cognition.conversation cc
+    JOIN
+        cognition.project cp
+    ON
+        cc.project_id = cp.id;"""
```
JWittmeyer: doesn't match our common sql style:

```
SELECT cc.*, cp.organization_id
FROM cognition.conversation cc
INNER JOIN cognition.project cp
    ON cc.project_id = cp.id
```
Also is there a specific reason why this isn't an orm sql alchemy function?

Also the xxx_with name usually indicates a filter. maybe `get_all_including_org_id` or `get_all_extend_org_id`

cognition_objects/conversation.py
```diff
@@ -25,6 +25,20 @@ def get(project_id: str, conversation_id: str) -> CognitionConversation:
     )
 
 
+def get_all_with_org() -> List[CognitionConversation]:
```
JWittmeyer: return type doesn't fit.

cognition_objects/conversation.py
LennartSchmidtKern: I tried it with sqlalchemy and always had to flatten the resulting org id with list comprehension

JWittmeyer (approved):

PR #144, opened 2024-11-26
Filter users only with email

lumburovskalina:
Main PR: https://github.com/code-kern-ai/refinery-gateway/pull/276

### DIFF
--- a/business_objects/user.py
+++ b/business_objects/user.py
@@ -157,6 +157,7 @@ def get_active_users_after_filter(
     FROM public.user u 
     LEFT JOIN organization o
         ON u.organization_id = o.id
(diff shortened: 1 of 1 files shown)


business_objects/user.py
```diff
@@ -157,6 +157,7 @@ def get_active_users_after_filter(
     FROM public.user u 
     LEFT JOIN organization o
         ON u.organization_id = o.id
+    WHERE u.email IS NOT NULL
```
JWittmeyer, line 160: with the condition below this will create a falty SQL statement (last_interaction needs an AND)

JWittmeyer (approved):

JWittmeyer (approved):

PR #149, opened 2024-12-10
Rework Factories

LennartSchmidtKern:
Main PR:
- [Rework Factories](https://github.com/code-kern-ai/cognition-gateway/pull/183)

### DIFF
--- a/cognition_objects/message.py
+++ b/cognition_objects/message.py
@@ -5,7 +5,6 @@ from ..session import session
 from ..models import CognitionMessage
 from ..util import prevent_sql_injection
 from .pipeline_version import get_current_version
-from collections import defaultdict
 
 
 def get_all_by_conversation_id(
@@ -25,19 +24,34 @@ def get_all_by_conversation_id(
 def get_all_by_conversation_ids(
     project_id: str, conversation_ids: List[str]
 ) -> Dict[str, List[CognitionMessage]]:
-    messages = (
-        session.query(CognitionMessage)
-        .filter(
-            CognitionMessage.project_id == project_id,
-            CognitionMessage.conversation_id.in_(conversation_ids),
-        )
-        .order_by(CognitionMessage.created_at.asc())
-        .all()
+
+    project_id = prevent_sql_injection(project_id, isinstance(project_id, str))
+    conversation_ids = prevent_sql_injection(
+        conversation_ids, isinstance(conversation_ids, list)
     )
-    messages_by_conversation = defaultdict(list)
-    for message in messages:
-        messages_by_conversation[message.conversation_id].append(message)
-    return messages_by_conversation
+    conversation_where = "conversation_id IN ('" + "','".join(conversation_ids) + "')"
+    query = f"""
+    SELECT jsonb_object_agg(conversation_id, messages)
+    FROM (
+        SELECT m.conversation_id,
+               array_agg(row_to_json(m)) AS messages
+        FROM (
+            SELECT *
+            FROM cognition.message
+            WHERE project_id = '{project_id}'
+            AND {conversation_where}
+            ORDER BY created_at ASC
+        ) m
+        GROUP BY m.conversation_id
+    ) x
+    """
+
+    message_info = general.execute_first(query)
+    if message_info and message_info[0]:
+        message_info = message_info[0]
+    else:
+        message_info = {}
+    return message_info
 
 
 def get_last_by_conversation_id(
--- a/cognition_objects/message.py
+++ b/cognition_objects/message.py
@@ -5,6 +5,7 @@ from ..session import session
 from ..models import CognitionMessage
 from ..util import prevent_sql_injection
 from .pipeline_version import get_current_version
+from collections import defaultdict
 
 
 def get_all_by_conversation_id(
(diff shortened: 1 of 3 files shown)


util.py
```diff
@@ -96,8 +96,14 @@ def sql_alchemy_to_dict(
...
     for_frontend: bool = False,
     column_whitelist: Optional[Iterable[str]] = None,
+    column_blacklist: Optional[Iterable[str]] = None,
+    column_rename_map: Optional[Dict[str, str]] = None,
 ):
-    result = __sql_alchemy_to_dict(sql_alchemy_object, column_whitelist)
+    result = __sql_alchemy_to_dict(
+        sql_alchemy_object, column_whitelist, column_blacklist
+    )
+    if column_rename_map:
```
JWittmeyer: this would only rename the topmost level. Not sure if intendet but probably not

util.py
```diff
@@ -96,8 +96,14 @@ def sql_alchemy_to_dict(
...
+        result = {column_rename_map.get(k, k): v for k, v in result.items()}
```
JWittmeyer: only works for dict like sql_alchemy_object changes. Could also be a list of dicts (e.g. an all request result)

util.py
LennartSchmidtKern: how far should it go, what is the second and third level?

JWittmeyer: we sometimes have list of dict with dicts inside. i think i would put it into the recouson so every level is changed if it has the provided key

cognition_objects/pipeline_log.py
```diff
@@ -220,3 +221,63 @@ def update_to_new_diff_structure(
...
+        )
+        .group_by(CognitionMessage.id)
+        .all()
+    )
+    grouped_logs_dict = {}
+    for message in grouped_message_logs:
+        grouped_logs_dict[message.id] = {
+            "time_logs_elapsed": message.time_logs_elapsed,
+            "logs_have_error": message.logs_have_error,
+        }
```
JWittmeyer: join should include a join criteria. With the added dict requirement i would recommend an sql query

```
SELECT jsonb_object_agg(id, jsonb_build_object('has_error',has_error, 'time_elapsed',time_elapsed))
FROM (
	SELECT m.id, SUM(pl.has_error::INT) > 0 has_error,SUM(pl.time_elapsed) time_elapsed
	FROM cognition.message m
	INNER JOIN cognition.pipeline_logs pl
		ON m.project_id = pl.project_id AND m.id = pl.message_id
	WHERE m.project_Id = '72c5721d-6cd4-4652-97b7-3aeca0600e07'
	-- AND m.conversation_id = '9fe61702-120e-4035-8b4b-e037ffb8812c'
	-- AND m.message_id = 'd5d547cc-e5ef-41e4-87f8-a53e578e6bcc'
	GROUP BY m.id 
)x
```

cognition_objects/pipeline_log.py
```diff
@@ -220,3 +221,63 @@ def update_to_new_diff_structure(
...
+    project_id: str,
+    conversation_ids: List[str],
+) -> Dict[str, CognitionPipelineLogs]:
+    grouped_message_logs = (
+        session.query(
+            CognitionMessage.id,
+            func.sum(CognitionPipelineLogs.time_elapsed).label("time_logs_elapsed"),
+            func.bool_or(CognitionPipelineLogs.has_error).label("logs_have_error"),
+        )
+        .join(CognitionMessage)
```
JWittmeyer: Not sure what is optimized by sqlalchemy but maybe print the full query to see how it would look (`general.print_orm_query`) because i suspect that the join with CognitionMessage on itself creates a query like
```
SELECT ...
FORM CognitionMessage
INNER JOIN CognitionMessage
INNER JOIN CognitionConversation
WHERE ...
```

which would be an unnecessary join on itself. The on criteria can btw be created like this:
image

cognition_objects/pipeline_log.py
```diff
@@ -220,3 +221,63 @@ def update_to_new_diff_structure(
 
     if with_commit:
         general.commit()
+
+
+def get_error_and_time_elapsed_by_conversation_id(
```
JWittmeyer: I'm not sure if we need two methods for basically the same code. 
I think conversation_ids can be Union[Iterable[str],str] and then in code with
```
if not isinstance(value, collections_abc_Iterable):
   conversation_ids = [conversation_ids]

```
or something similar

cognition_objects/message.py
```diff
@@ -21,6 +22,24 @@ def get_all_by_conversation_id(
...
+        .filter(
+            CognitionMessage.project_id == project_id,
+            CognitionMessage.conversation_id.in_(conversation_ids),
+        )
+        .order_by(CognitionMessage.created_at.asc())
+        .all()
+    )
+    messages_by_conversation = defaultdict(list)
+    for message in messages:
+        messages_by_conversation[message.conversation_id].append(message)
```
JWittmeyer: also doable in sql if we want to provide a dict instead of an sql alchemy object.

```
SELECT jsonb_object_agg(conversation_id, messages)
FROM (
	SELECT m.conversation_id, array_agg(row_to_json(M)) messages
	FROM cognition.message m 
	WHERE m.project_id = '72c5721d-6cd4-4652-97b7-3aeca0600e07'
	--AND m.conversation_id IN ('')
	GROUP BY m.conversation_id
) x
```

JWittmeyer (approved):

PR #153, opened 2025-02-11
ETL External API

andhreljaKern:
Main PR: https://github.com/code-kern-ai/cognition-gateway/pull/193

### DIFF
--- a/business_objects/personal_access_token_etl.py
+++ b/business_objects/personal_access_token_etl.py
@@ -63,7 +63,6 @@ def get_or_create(
             expires_at=expires_at,
             created_by=created_by,
         )
-        general.add(pat)
     return pat
 
 
@@ -85,10 +84,9 @@ def create(
     pat_scope = PersonalAccessTokenScopeEtl(
         scope=scope,
         subject=subject,
-        token_id=pat.id,
     )
-    general.add(pat_scope)
-    general.flush_or_commit(with_commit)
+    pat.scopes.append(pat_scope)
+    general.add(pat, with_commit)
     return pat
 
 
--- a/business_objects/personal_access_token_etl.py
+++ b/business_objects/personal_access_token_etl.py
@@ -1,5 +1,5 @@
 import datetime
-from typing import List
+from typing import List, Dict, Any
 from ..session import session
 from submodules.model.enums import TokenSubject
 from submodules.model.business_objects import general
@@ -21,7 +21,7 @@ INNER JOIN (
 """
 
 
-def get(token_id: str) -> CognitionPersonalAccessTokenEtl:
+def get(token_id: str) -> Dict[str, Any]:
     sql = TOKEN_WITH_SCOPE_QUERY + f"WHERE pate.id = '{token_id}'"
     return general.execute_first(sql)
 
@@ -40,12 +40,12 @@ def get_by_user_and_name(
     )
 
 
-def get_all(org_id: str) -> List[CognitionPersonalAccessTokenEtl]:
+def get_all(org_id: str) -> List[Dict[str, Any]]:
     sql = TOKEN_WITH_SCOPE_QUERY + f"WHERE pate.organization_id = '{org_id}'"
     return general.execute_all(sql)
 
 
-def get_all_by_user(user_id: str) -> List[CognitionPersonalAccessTokenEtl]:
+def get_all_by_user(user_id: str) -> List[Dict[str, Any]]:
     sql = TOKEN_WITH_SCOPE_QUERY + f"WHERE pate.created_by = '{user_id}'"
     return general.execute_all(sql)
 
--- a/business_objects/personal_access_token_etl.py
+++ b/business_objects/personal_access_token_etl.py
@@ -1,0 +1,103 @@
+import datetime
+from typing import List
+from ..session import session
+from submodules.model.business_objects import general
+from submodules.model.models import CognitionPersonalAccessTokenETL
+
+
+def get_by_user_and_name(
+    dataset_id: str,
+    created_by: str,
+    name: str,
+) -> CognitionPersonalAccessTokenETL:
+    return (
+        session.query(CognitionPersonalAccessTokenETL)
+        .filter(
+            CognitionPersonalAccessTokenETL.dataset_id == dataset_id,
+            CognitionPersonalAccessTokenETL.name == name,
+            CognitionPersonalAccessTokenETL.created_by == created_by,
+        )
+        .first()
+    )
+
+
+def get_all(dataset_id: str) -> List[CognitionPersonalAccessTokenETL]:
+    return (
+        session.query(CognitionPersonalAccessTokenETL)
+        .filter(CognitionPersonalAccessTokenETL.dataset_id == dataset_id)
+        .all()
+    )
+
+
+def get_by_token(dataset_id: str, token: str) -> CognitionPersonalAccessTokenETL:
+    return (
+        session.query(CognitionPersonalAccessTokenETL)
+        .filter(
+            CognitionPersonalAccessTokenETL.dataset_id == dataset_id,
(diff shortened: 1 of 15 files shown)


util.py
```diff
@@ -145,12 +145,27 @@ def rename_columns(data: Any) -> Any:
         }
         return rename_columns(result)
     elif isinstance(sql_alchemy_object, Base):
-        result = {
-            c.name: getattr(sql_alchemy_object, c.name)
-            for c in sql_alchemy_object.__table__.columns
-            if (not column_whitelist or c.name in column_whitelist)
-            and (not column_blacklist or c.name not in column_blacklist)
-        }
+        table_args = getattr(sql_alchemy_object, "__table_args__", {})
```
andhreljaKern: I was trying to find a way to default the columns to display because `__table__.columns` omit "relationship" columns. Until this change, I couldn't access `CognitionPersonalAccessTokenEtl(**).scopes`. Is there somehing we already have built-in for this use case? 

I am open to all suggestions because I think this should be much more simple

models.py
```diff
@@ -1,37 +1,39 @@
 import uuid
 
 from .enums import (
```
andhreljaKern, line 3: sort model imports asc

util.py
JWittmeyer: it's intentional. The parent_to_child relationships shouldn't be used anymore. They are old logic from gql but lead to often miss use in code. If it's needed we usually write a query with an itentianal join to then be used as an object directly. i can show you how :)

business_objects/personal_access_token_etl.py
```diff
@@ -0,0 +1,136 @@
...
+        session.query(CognitionPersonalAccessTokenEtl)
+        .filter(
+            CognitionPersonalAccessTokenEtl.name == name,
+            CognitionPersonalAccessTokenEtl.created_by == created_by,
+        )
+        .first()
+    )
+
+
+def get_all(user_id: str) -> List[CognitionPersonalAccessTokenEtl]:
```
JWittmeyer: get_all_by_user_id

business_objects/personal_access_token_etl.py
```diff
@@ -0,0 +1,136 @@
...
+) -> CognitionPersonalAccessTokenEtl:
+    pat = get_by_user_and_name(created_by, name)
+    if not pat:
+        pat = CognitionPersonalAccessTokenEtl(
+            name=name,
+            token=token,
+            expires_at=expires_at,
+            created_by=created_by,
+        )
+    return pat
```
JWittmeyer: auto create shouldn't be used. also would miss a commit if it's called alone

business_objects/personal_access_token_etl.py
```diff
@@ -0,0 +1,136 @@
...
+        name=name,
+        token=token,
+        expires_at=expires_at,
+        created_by=created_by,
+    )
+    pat_scope = PersonalAccessTokenScopeEtl(
+        scope=scope,
+        subject=subject,
+    )
+    pat.scopes.append(pat_scope)
```
JWittmeyer, line 116: we don't usually use parent to child relationships anymore

models.py
```diff
@@ -1428,6 +1430,56 @@ class CognitionPersonalAccessToken(Base):
...
+        UUID(as_uuid=True),
+        ForeignKey(f"{Tablenames.USER.value}.id", ondelete="SET NULL"),
+        index=True,
+    )
+    created_at = Column(DateTime, default=sql.func.now())
+    name = Column(String)
+    expires_at = Column(DateTime)
+    last_used = Column(DateTime)
+    token = Column(String)
+    scopes = parent_to_child_relationship(
```
JWittmeyer, line 1452: old logic from gql shouldn't be used anymore

models.py
```diff
@@ -1428,6 +1430,56 @@ class CognitionPersonalAccessToken(Base):
...
+
+class PersonalAccessTokenScopeEtl(Base):
+    __tablename__ = Tablenames.PERSONAL_ACCESS_TOKEN_SCOPE_ETL.value
+    __table_args__ = {"schema": "cognition"}
+    id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    created_at = Column(DateTime, default=sql.func.now())
+    scope = Column(String, default=TokenScope.READ_WRITE.value)
+    subject = Column(String, default=TokenSubject.PROJECT.value)
+    subject_id = Column(UUID(as_uuid=True))  # project_id or dataset_id
+    token_id = Column(
```
JWittmeyer, line 1466: afaik the idea was to later also use this table for non etl tokens wich would make this fkey relation problematic

models.py
```diff
@@ -1428,6 +1430,56 @@ class CognitionPersonalAccessToken(Base):
...
+            "scopes",
+        ],
+    }
+    id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    created_by = Column(
+        UUID(as_uuid=True),
+        ForeignKey(f"{Tablenames.USER.value}.id", ondelete="SET NULL"),
+        index=True,
+    )
+    created_at = Column(DateTime, default=sql.func.now())
```
JWittmeyer, line 1447: should include org_id since our SOs switch orgs

business_objects/personal_access_token_etl.py
andhreljaKern: this is copy-pasted from personal_access_token.py

business_objects/personal_access_token_etl.py
andhreljaKern: I only introduced this get_or_create because an existing token can have multiple scopes and I wanted to avoid adding new "create scope" endpoints. This function should only be called via the `create` function. To reduce ambiguity, I can rename `get_or_create` into `__get_or_create` + add `with_commit` argument + `general.add`

If we were to refactor other PAT to use scopes, this would have to be rewritten so that `token_scope` endpoints are added.

business_objects/personal_access_token_etl.py
andhreljaKern: I saw that, I don't really understand why, I was having difficulties developing some features because there wasn't a parent-child relationship between the entities

models.py
andhreljaKern: yeah, if we were to use project tokens with this structure, we would need to perform a data migration (or ask users to recreate their tokens - this is a valid vendor request IMO - with the new tokens, they also wouldn't need a token for each project they need to access). The fkey relation is created to allow tokens to be easily reused

business_objects/personal_access_token_etl.py
JWittmeyer: works for me

business_objects/personal_access_token_etl.py
andhreljaKern: get_all with org_id

util.py
```diff
@@ -146,7 +146,7 @@ def rename_columns(data: Any) -> Any:
         return rename_columns(result)
     elif isinstance(sql_alchemy_object, Base):
         result = {
-            c.name: getattr(sql_alchemy_object, c.name)
+            c.name: __sql_alchemy_to_dict(getattr(sql_alchemy_object, c.name))
```
andhreljaKern: @JWittmeyer should we keep this change?

JWittmeyer: no

models.py
```diff
@@ -1428,6 +1430,61 @@ class CognitionPersonalAccessToken(Base):
...
+        "include_columns": [
+            "id",
+            "created_by",
+            "created_at",
+            "name",
+            "expires_at",
+            "last_used",
+            "token",
+            "scopes",
+        ],
```
JWittmeyer: is this still needed?

andhreljaKern: not needed, removed

JWittmeyer (approved):

PR #160, opened 2025-03-24
Qdrant results limit

anmarhindi:
Link to main PR:
- https://github.com/code-kern-ai/refinery-neural-search/pull/82

### DIFF
--- a/business_objects/embedding.py
+++ b/business_objects/embedding.py
@@ -441,6 +441,8 @@ def get_match_record_ids_to_qdrant_ids_with_max_score(
                 project_id, embedding_id, chunk, limit
             ):
                 results.extend(r)
+        if len(results) > limit:
+            results = results[:limit]
         return results
 
     project_id = prevent_sql_injection(project_id, isinstance(project_id, str))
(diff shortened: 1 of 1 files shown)


business_objects/embedding.py
```diff
@@ -441,6 +441,8 @@ def get_match_record_ids_to_qdrant_ids_with_max_score(
                 project_id, embedding_id, chunk, limit
             ):
                 results.extend(r)
+        if len(results) > limit:
```
andhreljaKern: if can be omitted, `[1,2,3][:6]` works without issues

andhreljaKern: also, should we break the for-loop if results are extended over the limit?

anmarhindi: Regarding the first point, we **could** add it, however, the limit is not guaranteed to be above or equal to 0, which might change the behaviour if we get a -1 or something. This is not enforced, hence the if is safer as len is never below 0.

However, I can introduce a max that will check for this and make it cleaner

For the second point, yes I will add a break for it.

andhreljaKern (approved):

PR #165, opened 2025-04-24
Added table for full admin access

lumburovskalina:
Main PR: https://github.com/code-kern-ai/admin-dashboard/pull/75

### DIFF
--- a/business_objects/user.py
+++ b/business_objects/user.py
@@ -182,6 +182,7 @@ def update_last_interaction(user_id: str) -> None:
 
 
 def check_email_in_full_admin(email: str) -> bool:
+    email = email.lower()
     email = prevent_sql_injection(email, isinstance(email, str))
     query = f"""
     SELECT EXISTS (
@@ -190,4 +191,7 @@ def check_email_in_full_admin(email: str) -> bool:
         WHERE email = '{email}'
     )
     """
-    return general.execute_first(query)[0]
+    result = general.execute_first(query)
+    if result and result[0]:
+        return True
+    return False
--- a/business_objects/user.py
+++ b/business_objects/user.py
@@ -179,3 +179,15 @@ def update_last_interaction(user_id: str) -> None:
     user_item = get(user_id)
     user_item.last_interaction = sql.func.now()
     general.commit()
+
+
+def check_email_in_full_admin(email: str) -> bool:
+    email = prevent_sql_injection(email, isinstance(email, str))
+    query = f"""
+    SELECT EXISTS (
+        SELECT 1
+        FROM global.full_admin_access
+        WHERE email = '{email}'
+    )
+    """
+    return general.execute_first(query)[0]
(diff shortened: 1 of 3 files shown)


business_objects/user.py
```diff
@@ -179,3 +179,15 @@ def update_last_interaction(user_id: str) -> None:
...
+def check_email_in_full_admin(email: str) -> bool:
+    email = prevent_sql_injection(email, isinstance(email, str))
+    query = f"""
+    SELECT EXISTS (
+        SELECT 1
+        FROM global.full_admin_access
+        WHERE email = '{email}'
+    )
+    """
+    return general.execute_first(query)[0]
```
JWittmeyer: doesn't actually return a boolean, should be converted (

```
result = general.execute_first(query)
if result and result[0]:
    return True
return False
```

business_objects/user.py
```diff
@@ -179,3 +179,15 @@ def update_last_interaction(user_id: str) -> None:
     user_item = get(user_id)
     user_item.last_interaction = sql.func.now()
     general.commit()
+
+
+def check_email_in_full_admin(email: str) -> bool:
+    email = prevent_sql_injection(email, isinstance(email, str))
```
JWittmeyer, line 186: since emails are stored lower case we should also check lower case

```suggestion
    email = email.lower()
    email = prevent_sql_injection(email, isinstance(email, str))
```

models.py
```diff
@@ -2062,3 +2062,11 @@ class PlaygroundQuestion(Base):
...
     # record_ids = Column(JSON)
     # meta_info = Column(JSON)
+
+
+class FullAdminAccess(Base):
+    __tablename__ = Tablenames.FULL_ADMIN_ACCESS.value
+    __table_args__ = {"schema": "global"}
+    id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    email = Column(String, unique=True)
+    meta_info = Column(JSON)
```
JWittmeyer, line 2072: what is this for?

lumburovskalina: for now it is not used, but i added it for any additional configuration if we need it in future

andhreljaKern: IMO we should be using a different approach here. Instead of the `full_admin_access` table, we should create a `user_role (id, name)` table and expand the `user` model to include `user_role_id_fk`. Role name should, for now, contain values `user` and `admin` (additionally maybe `superuser` for even more elevated permissions). We could default all existing users to the role `user` and only set `admin` or `superuser` manually (or through admin-dashboard).

This way, we can further expand on permissions in the future by adding a new table `user_role_permissions (id, user_role_id_fk, permission_target, permission, permission_target_id NULLABLE)` with e.g. record `(<uuid>, <uuid>, 'refinery_project', 'read_write', <refinery_project uuid>)`.

JWittmeyer: for now we keep it as currently implemented. We need to rework the role system at some point for the whole application and a partial but more complex solution will only add to the complexity when we have the change to implement. For now a simple lookup in a table is sufficient

JWittmeyer (approved):

PR #166, opened 2025-05-06
Adds cache decorator & session decorator

JWittmeyer:
main PR: https://github.com/code-kern-ai/cognition-gateway/pull/215

### DIFF
--- a/db_cache.py
+++ b/db_cache.py
@@ -81,13 +81,13 @@ class TTLCacheDecorator:
                 if entry:
                     value, expires_at = entry
                     if now < expires_at:
-                        print(f"Cache hit for {key} in {self.cache_type}")
+                        # print(f"Cache hit for {key} in {self.cache_type}")
                         return value
                     # expired
                     del cache[key]
 
             # miss or expired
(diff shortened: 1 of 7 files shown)


db_cache.py
```diff
@@ -0,0 +1,180 @@
...
+                        print(f"Cache hit for {key} in {self.cache_type}")
+                        return value
+                    # expired
+                    del cache[key]
+
+            # miss or expired
+            print(f"No cache hit for {key} in {self.cache_type}")
+            result = fn(*args, **kwargs)
+            with _CACHE_LOCK:
+                cache = _GLOBAL_CACHE_MAP[self.cache_type]
```
andhreljaKern, line 94: does the wrapper work with multiple logged in users? I understand this is the cache map definition:
```python
_GLOBAL_CACHE_MAP = {
    "user": "<user-uuid>"
}
```
is this dictionary recreated per each logged in user?

JWittmeyer: it's actually 
```
{
"user": { "<user_id>": Cached object}
}
```

so it should work (hopefully) - the test will show us more :D

PR #167, opened 2025-05-15
Cognition integration provider

andhreljaKern:
Main PR: https://github.com/code-kern-ai/refinery-gateway/pull/302


business_objects/project.py
```diff
@@ -609,3 +609,14 @@ def get_project_by_project_id_sql(project_id: str) -> Dict[str, Any]:
         return value[0]
     else:
         return None
+
+
+def get_by_name_and_org_id(name: str, organization_id: str) -> Optional[Project]:
```
JWittmeyer: what is this used for? Getting by name is always tricky. we in theory should have unique names but there is not really a guarantee since the snapshot import doesn't force it:
image


If possible we should use id's for linking

cognition_objects/integration.py
```diff
@@ -0,0 +1,254 @@
...
+        session.query(CognitionIntegration)
+        .filter(CognitionIntegration.id == id)
+        .first()
+    )
+
+
+def get_all(
+    integration_type: Optional[str] = None,
+    exclude_failed: Optional[bool] = False,
+    only_synced: Optional[bool] = False,
```
JWittmeyer: optional is only for allowing None values, if it's either True or False then `bool = False` is used

cognition_objects/integration.py
```diff
@@ -0,0 +1,254 @@
...
+        )
+    if only_synced:
+        query = query.filter(CognitionIntegration.is_synced == True)
+    return query.order_by(CognitionIntegration.created_at.desc()).all()
+
+
+def get_all_in_org(
+    org_id: str,
+    integration_type: Optional[str] = None,
+    only_synced: Optional[bool] = False,
```
JWittmeyer: see above

cognition_objects/integration.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def get_all_in_org_paginated(
+    org_id: str,
+    integration_type: Optional[str] = None,
+    page: int = 1,
+    page_size: int = 10,
+) -> List[CognitionIntegration]:
+    schema_name = CognitionIntegration.__table__.schema or "public"
+    table_name = f"{schema_name}.{CognitionIntegration.__tablename__}"
```
JWittmeyer: we usually dont use the schema / table name dynamically but write it in the sql statement directly.

cognition_objects/integration.py
```diff
@@ -0,0 +1,254 @@
...
+        CognitionIntegration.organization_id == org_id
+    )
+    if integration_type:
+        query = query.filter(CognitionIntegration.type == integration_type)
+    if only_synced:
+        query = query.filter(CognitionIntegration.is_synced == True)
+    return query.order_by(CognitionIntegration.created_at.desc()).all()
+
+
+def get_all_in_org_paginated(
```
JWittmeyer, line 68: where is this function used? couldn't find it in cognition integration provider

cognition_objects/integration.py
```diff
@@ -0,0 +1,254 @@
...
+    sql = f"""
+    SELECT id FROM (
+        SELECT 
+            ROW_NUMBER () OVER(PARTITION BY intg.id ORDER BY intg.created_at ASC) rn,
+            intg.id
+        FROM {table_name} intg
+        WHERE intg.organization_id = '{org_id}'
+    ) pages
+    WHERE rn BETWEEN {first_page} AND {last_page}
+    """
```
JWittmeyer: this feels like a uncommon way to collect paginated results.
especially since we paginate on the full query but later filter by type. potentially resulting in 0 values.

Usually an sql alchemy orm query with corresponding filter + limit & offset do the trick.

cognition_objects/integration.py
```diff
@@ -0,0 +1,254 @@
...
+    if started_at is not None:
+        integration.started_at = started_at
+    if last_synced_at is not None:
+        integration.last_synced_at = last_synced_at
+    if delta_criteria is not None:
+        integration.delta_criteria = delta_criteria
+        flag_modified(integration, "delta_criteria")
+
+    integration.is_synced = is_synced
+    integration.finished_at = finished_at
```
JWittmeyer: if we update e.g. the state this automatically sets the is_synced & finished_at to None. I'm not sure this is intended. (e.g. for an update of the llm config we dont need to resync)

If the None is the value to be set in some instances we usually work around it by using a "NULL" keyword or similar

cognition_objects/integration.py
```diff
@@ -0,0 +1,254 @@
...
+def execution_finished(id: str) -> bool:
+    return bool(
+        session.query(CognitionIntegration)
+        .filter(
+            CognitionIntegration.id == id,
+            CognitionIntegration.state.in_(
+                [
+                    CognitionMarkdownFileState.FINISHED.value,
+                    CognitionMarkdownFileState.FAILED.value,
+                ]
```
JWittmeyer: this creates the array on every call of the function. outside as const set FINISHED_STATES would be preferred

cognition_objects/integration.py
```diff
@@ -0,0 +1,254 @@
...
+def delete_many(
+    ids: List[str], delete_refinery_projects: bool = False, with_commit: bool = True
+) -> None:
+    integrations = session.query(CognitionIntegration).filter(
+        CognitionIntegration.id.in_(ids)
+    )
+    if delete_refinery_projects:
+        session.query(Project).filter(
+            Project.id.in_(filter(None, [i.project_id for i in integrations]))
+        ).delete(synchronize_session=False)
```
JWittmeyer: as notes in direct comments this only removes from the db not the corresponding minio data. or similar

cognition_objects/integration_access.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def get(
+    org_id: str, integration_type: CognitionIntegrationType
+) -> List[CognitionIntegrationAccess]:
+    return (
+        session.query(CognitionIntegrationAccess)
+        .filter(
+            CognitionIntegrationAccess.organization_id == org_id,
+            CognitionIntegrationAccess.integration_type == integration_type,
```
JWittmeyer: .value missing or wrong type used in function

enums.py
```diff
@@ -871,3 +884,88 @@ class EvaluationRunState(Enum):
...
     SUCCESS = "SUCCESS"
     FAILED = "FAILED"
+
+
+class CognitionIntegrationType(Enum):
+    # CSV = "CSV"
+    # JSON = "JSON"
+    # DOCX = "DOCX"
+    # XLSX = "XLSX"
+    # WEBPAGE = "WEBPAGE"
```
JWittmeyer: remove

enums.py
```diff
@@ -871,3 +884,88 @@ class EvaluationRunState(Enum):
...
+        changed_value = value.upper().replace(" ", "_").replace("-", "_")
+        try:
+            return CognitionIntegrationType[changed_value]
+        except KeyError:
+            raise KeyError(
+                f"Could not parse CognitionIntegrationType from string '{changed_value}'"
+            )
+
+
+class IntegrationMetadata(Enum):
```
JWittmeyer: this looks for like a helper file/class than an enum

integration_objects/manager.py
```diff
@@ -0,0 +1,220 @@
...
+        )
+        .order_by(IntegrationModel.created_at.asc())
+        .all()
+    )
+
+
+def get_existing_integration_records(
+    IntegrationModel: Type,
+    integration_id: str,
+    by: Optional[str] = "source",
```
JWittmeyer: type `str = "source"`

integration_objects/manager.py
```diff
@@ -0,0 +1,220 @@
...
+    return {
+        getattr(record, by, record.source): record
+        for record in get_all_by_integration_id(IntegrationModel, integration_id)
+    }
+
+
+def get_running_ids(
+    IntegrationModel: Type,
+    integration_id: str,
+    by: Optional[str] = "source",
```
JWittmeyer: see above

integration_objects/manager.py
```diff
@@ -0,0 +1,220 @@
...
+) -> int:
+    return dict(
```
JWittmeyer: return type & return dont seem to fit

integration_objects/manager.py
```diff
@@ -0,0 +1,220 @@
...
+                f"Invalid field '{key}' for {IntegrationModel.__tablename__}"
+            )
+        existing_value = getattr(integration_record, key, None)
+        if value is not None and value != existing_value:
+            setattr(integration_record, key, value)
+            flag_modified(integration_record, key)
+            record_updated = True
+
+    if record_updated:
+        general.add(integration_record, with_commit=with_commit)
```
JWittmeyer: `general.flush_or_commit(with_commit)` probably

integration_objects/manager.py
```diff
@@ -0,0 +1,220 @@
...
+
+def clear_history(
+    IntegrationModel: Type,
+    id: str,
+    with_commit: bool = False,
+) -> None:
+    integration_record = get_by_id(IntegrationModel, id)
+    integration_record.delta_criteria = None
+    flag_modified(integration_record, "delta_criteria")
+    general.add(integration_record, with_commit)
```
JWittmeyer: flush_or_commit

models.py
```diff
@@ -2077,3 +2077,245 @@ class FullAdminAccess(Base):
...
+    project_id = Column(
+        UUID(as_uuid=True),
+        ForeignKey(f"{Tablenames.PROJECT.value}.id", ondelete="CASCADE"),
+        index=True,
+        nullable=True,
+    )
+    created_by = Column(
+        UUID(as_uuid=True),
+        ForeignKey(f"{Tablenames.USER.value}.id", ondelete="SET NULL"),
+        index=False,
```
JWittmeyer: all our f-keys are indexed

models.py
```diff
@@ -2077,3 +2077,245 @@ class FullAdminAccess(Base):
...
+    )
+    created_at = Column(DateTime, default=sql.func.now())
+    updated_by = Column(
+        UUID(as_uuid=True),
+        ForeignKey(f"{Tablenames.USER.value}.id", ondelete="SET NULL"),
+        index=False,
```
JWittmeyer: see above

models.py
```diff
@@ -2077,3 +2077,245 @@ class FullAdminAccess(Base):
...
+
+
+class CognitionIntegrationAccess(Base):
+    __tablename__ = Tablenames.INTEGRATION_ACCESS.value
+    __table_args__ = {"schema": "cognition"}
+    id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    created_by = Column(
+        UUID(as_uuid=True),
+        ForeignKey(f"{Tablenames.USER.value}.id", ondelete="SET NULL"),
+        index=False,
```
JWittmeyer: see above

models.py
```diff
@@ -2077,3 +2077,245 @@ class FullAdminAccess(Base):
...
+            "source",
+            name=f"unique_{__tablename__}_source",
+        ),
+        {"schema": "integration"},
+    )
+    id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    created_by = Column(
+        UUID(as_uuid=True),
+        ForeignKey(f"{Tablenames.USER.value}.id", ondelete="SET NULL"),
+        index=False,
```
JWittmeyer: see above

models.py
```diff
@@ -2077,3 +2077,245 @@ class FullAdminAccess(Base):
...
+    )
+    updated_by = Column(
+        UUID(as_uuid=True),
+        ForeignKey(f"{Tablenames.USER.value}.id", ondelete="SET NULL"),
+        index=False,
+        nullable=True,
```
JWittmeyer: see above & nullable= True is afaik default (otherwise the cascade wouldn't work)

models.py
```diff
@@ -2077,3 +2077,245 @@ class FullAdminAccess(Base):
...
```
JWittmeyer: see above

models.py
```diff
@@ -2077,3 +2077,245 @@ class FullAdminAccess(Base):
...
+    )
+    created_at = Column(DateTime, default=sql.func.now())
+    updated_at = Column(DateTime, default=None, onupdate=sql.func.now())
```
JWittmeyer: i'm unfamiliar with the onupdate call. Is this also set via a normal commit?

models.py
```diff
@@ -2077,3 +2077,245 @@ class FullAdminAccess(Base):
...
+            "source",
+            name=f"unique_{__tablename__}_source",
+        ),
+        {"schema": "integration"},
+    )
+    id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
+    created_by = Column(
+        UUID(as_uuid=True),
+        ForeignKey(f"{Tablenames.USER.value}.id", ondelete="SET NULL"),
+        index=False,
```
JWittmeyer: index

models.py
```diff
@@ -2077,3 +2077,245 @@ class FullAdminAccess(Base):
...
+    )
+    updated_by = Column(
+        UUID(as_uuid=True),
+        ForeignKey(f"{Tablenames.USER.value}.id", ondelete="SET NULL"),
+        index=False,
+        nullable=True,
```
JWittmeyer: index

util.py
```diff
@@ -208,6 +208,13 @@ def to_camel_case(name: str):
     return "".join([name[0].lower(), name[1:]])
 
 
+def to_snake_case(name: str):
+    if not is_camel_case(name):
+        return name
+    name = sub("(.)([A-Z][a-z]+)", r"\1_\2", name)
+    return sub("([a-z0-9])([A-Z])", r"\1_\2", name).lower()
```
JWittmeyer: regex pattern precompiled outside the function

business_objects/project.py
andhreljaKern: This is a legacy function used by cognition-gateway, removed

cognition_objects/integration.py
andhreljaKern: used by cognition-gateway

cognition_objects/integration.py
andhreljaKern: that makes sense, these values are consumed by the UI. Value _is_synced_ `null` is a signal for the "Check for updates" spinner. Value _finished_at_ is a signal for the FileState switches (extracting, splitting, transforming...). Is it safe for me to use value `'NULL'` instead of `null`?

enums.py
andhreljaKern: what about having it as a `submodules.model.integration_objects.enums`?

models.py
andhreljaKern: as far as I understand from https://docs.sqlalchemy.org/en/20/core/defaults.html#python-executed-functions (and looking at the DB), it is automatically set by sqlalchemy. We do not provide this value manually via `integration_db_bo.update`

cognition_objects/integration.py
andhreljaKern: please see https://github.com/code-kern-ai/refinery-gateway/pull/302#issuecomment-3000473132

util.py
andhreljaKern: does this work? ([reference](https://stackoverflow.com/questions/1175208/elegant-python-function-to-convert-camelcase-to-snake-case))
```python
SNAKE_CASE_PATTERNS = [
    compile(r"(.)([A-Z][a-z]+)"),
    compile(r"([a-z0-9])([A-Z])"),
]
def to_snake_case(name: str):
    if not is_camel_case(name):
        return name
    for phase in SNAKE_CASE_PATTERNS:
        name = phase.sub(r"\1_\2", name)
    return name.lower()
```

cognition_objects/integration.py
andhreljaKern: updated to:

```python
def get_all_in_org_paginated(
    org_id: str,
    integration_type: Optional[str] = None,
    page: int = 1,
    page_size: int = 10,
) -> List[CognitionIntegration]:
    query = session.query(CognitionIntegration).filter(
        CognitionIntegration.organization_id == org_id,
    )

    if integration_type:
        query = query.filter(CognitionIntegration.type == integration_type)

    return (
        query.order_by(CognitionIntegration.created_at.desc())
        .limit(page_size)
        .offset((page - 1) * page_size)
        .all()
    )

```

cognition_objects/integration.py
andhreljaKern:
```python
if is_synced is not None:
    if is_synced == 'NULL':
        integration.is_synced = None
    else:
        integration.is_synced = is_synced
```

cognition_objects/integration.py
andhreljaKern: `.offset(max(0, (page - 1) * page_size))`

enums.py
andhreljaKern: try to use it as a standalone helper object `submodules.model.integration_objects.helper`

business_objects/monitor.py
```diff
@@ -197,6 +198,16 @@ def set_parse_cognition_file_task_to_failed(
     general.commit()
 
 
+def set_integration_task_to_failed(
```
andhreljaKern, line 202: should be moved to cognition_objects/integration.py

business_objects/project.py
```diff
@@ -155,6 +152,39 @@ def get_all(organization_id: str) -> List[Project]:
...
 
 
+def get_all_with_access_management(organization_id: str) -> List[Project]:
+    return (
+        session.query(Project)
+        .join(Attribute, Project.id == Attribute.project_id)
+        .filter(
+            Project.organization_id == organization_id,
+            Attribute.name.in_(["__ACCESS_GROUPS", "__ACCESS_USERS"]),  #
+            Attribute.name.in_(["__ACCESS_GROUPS", "__ACCESS_USERS"]),
```
andhreljaKern: should one of these 2 filters be removed?

business_objects/monitor.py
```diff
@@ -197,6 +198,16 @@ def set_parse_cognition_file_task_to_failed(
...
+    integration_id: str,
+    with_commit: bool = False,
+) -> None:
+    integration = integration_db_bo.get_by_id(integration_id)
+    if integration:
+        integration.state = enums.CognitionMarkdownFileState.FAILED.value
```
andhreljaKern:
```suggestion
        integration_db_bo.update(
            integration_id,
            state=enums.CognitionMarkdownFileState.FAILED.value,
            finished_at=datetime.now(datetime.timezone.utc),
            is_synced=False,
            error_message=?
        )
```

business_objects/record.py
```diff
@@ -807,6 +807,25 @@ def delete_user_created_attribute(
...
+        record_items = get_all(project_id=project_id)
+        for i, record_item in enumerate(record_items):
+            if record_item.data.get(access_groups_attribute_item.name):
+                del record_item.data[access_groups_attribute_item.name]
+            if record_item.data.get(access_users_attribute_item.name):
+                del record_item.data[access_users_attribute_item.name]
+            flag_modified(record_item, "data")
+            if (i + 1) % 1000 == 0:
+                general.flush_or_commit(with_commit)
+        general.flush_or_commit(with_commit)
```
andhreljaKern, line 834: why commit every time, twice? Is it because the process takes so long the session may break during that time?

business_objects/record.py
```diff
@@ -925,3 +944,19 @@ def get_first_no_text_column(project_id: str, record_id: str) -> str:
     WHERE r.project_id = '{project_id}' AND r.id = '{record_id}'
     """
     return general.execute_first(query)[0]
```
andhreljaKern, line 954: typing

cognition_objects/group.py
```diff
@@ -0,0 +1,123 @@
...
+        .all()
+    )
+    integration_groups_by_permission = {}
+    for group in integration_groups:
+        permission_id = group.meta_data.get("permission_id")
+        integration_groups_by_permission[permission_id] = group
+    return integration_groups_by_permission
+
+
+def get_by_name(organization_id: str, name: str):
```
andhreljaKern: typing

cognition_objects/group_member.py
```diff
@@ -0,0 +1,103 @@
+from datetime import datetime
+from typing import Optional
+from ..business_objects import general, user
+from . import group
+from ..session import session
+from ..models import CognitionGroupMember
+
+
+def get(group_id: str, id: str):
```
andhreljaKern: typing

cognition_objects/group_member.py
```diff
@@ -0,0 +1,103 @@
...
+        session.query(CognitionGroupMember)
+        .filter(
+            CognitionGroupMember.group_id == group_id,
+            CognitionGroupMember.user_id == user_id,
+        )
+        .first()
+    )
+
+
+def get_by_user_id(user_id: str) -> list:
```
andhreljaKern:
```suggestion
def get_by_user_id(user_id: str) -> List[CognitiongGroupMember]:
```

cognition_objects/group_member.py
```diff
@@ -0,0 +1,103 @@
...
+    return (
+        session.query(CognitionGroupMember)
+        .filter(CognitionGroupMember.user_id == user_id)
+        .all()
+    )
+
+
+def get_all_by_group(group_id: str) -> list:
```
andhreljaKern:
```suggestion
def get_all_by_group(group_id: str) -> List[CognitionGroupMember]:
```

integration_objects/helper.py
```diff
@@ -0,0 +1,52 @@
...
+        "description",
+        "size",
+        "mime_type",
+        "hashes",
+        "permissions",
+    },
+}
+
+
+def get_supported_metadata_keys(table_name: str):
```
andhreljaKern:
```suggestion
def get_supported_metadata_keys(table_name: str) -> set[str]:
```

integration_objects/manager.py
```diff
@@ -0,0 +1,225 @@
...
+) -> None:
+    integration_record = get_by_id(IntegrationModel, id)
+    integration_record.delta_criteria = None
+    flag_modified(integration_record, "delta_criteria")
+    general.flush_or_commit(with_commit)
+
+
+def get_supported_metadata(
+    table_name: str, metadata: Dict[str, Union[str, int, float, bool]]
+) -> None:
```
andhreljaKern:
```suggestion
) -> Dict[str, Any]:
```

integration_objects/manager.py
```diff
@@ -0,0 +1,225 @@
+from typing import List, Optional, Dict, Union, Type
```
andhreljaKern:
```suggestion
from typing import List, Optional, Dict, Union, Type, Any
```

integration_objects/manager.py
```diff
@@ -0,0 +1,225 @@
...
+    supported_keys = get_supported_metadata_keys(table_name)
+    supported_metadata = {
+        key: metadata[key] for key in supported_keys.intersection(metadata.keys())
+    }
+    return __rename_metadata(table_name, supported_metadata)
+
+
+def __rename_metadata(
+    table_name: str, metadata: Dict[str, Union[str, int, float, bool]]
+) -> Dict[str, object]:
```
andhreljaKern:
```suggestion
) -> Dict[str, Any]:
```

util.py
```diff
@@ -216,6 +220,15 @@ def to_camel_case(name: str):
     return "".join([name[0].lower(), name[1:]])
 
 
+def to_snake_case(name: str):
```
andhreljaKern:
```suggestion
def to_snake_case(name: str) -> str:
```

business_objects/project.py
```diff
@@ -155,6 +153,62 @@ def get_all(organization_id: str) -> List[Project]:
...
+    query = f"""
+    SELECT DISTINCT
+            p.*,
+            CASE
+                WHEN
+                    ci.id IS NOT NULL
+                    AND (ci.config -> 'extract_kwargs' ->> 'sync_sharepoint_permissions')::text = 'true'
+                THEN TRUE
+                ELSE FALSE
+            END AS is_sharepoint_sync_active
```
andhreljaKern: Given there are multiple ways we infer this boolean (https://github.com/code-kern-ai/cognition-gateway/pull/225#discussion_r2171130448), maybe it's best to have it as a project table column? cc. @JWittmeyer

JWittmeyer: I wouldn't create a new column for it but i would change the sql code to be easier readable. 
Not sure if this is meant by it but maybe this does the trick `COALESCE((ci.config -> 'extract_kwargs' ->> 'sync_sharepoint_permissions')::BOOLEAN,FALSE) AS is_sharepoint_sync_active`
Example:
```
SELECT
  COALESCE((b->'test'->>'bool')::BOOLEAN,FALSE) AS val
FROM (
  SELECT json_build_object('test',json_build_object('bool', TRUE)) AS b
  UNION ALL SELECT json_build_object('test',json_build_object('bool', FAlse))
  UNION ALL SELECT json_build_object('test',NULL)
  UNION ALL SELECT NULL
) x;

```

business_objects/record.py
LennartSchmidtKern: I took this pattern from "delete_user_created_attribute"

business_objects/record.py
LennartSchmidtKern: not part of PR?

cognition_objects/group.py
```diff
@@ -0,0 +1,127 @@
...
+        .filter(
+            CognitionGroup.organization_id == organization_id,
+            integration_id_json == integration_id,
+        )
+        .all()
+    )
+    integration_groups_by_permission = {}
+    for group in integration_groups:
+        permission_id = group.meta_data.get("permission_id")
+        integration_groups_by_permission[permission_id] = group
```
JWittmeyer:
```suggestion
    integration_groups_by_permission = {group.meta_data.get("permission_id"):group for group in integration_groups}

```

integration_objects/manager.py
```diff
@@ -0,0 +1,235 @@
...
+from ..cognition_objects import integration as integration_db_bo
+from ..session import session
+from .helper import get_supported_metadata_keys
+
+
+def get(
+    IntegrationModel: Type,
+    integration_id: str,
+    id: Optional[str] = None,
+) -> object:
```
JWittmeyer: Union[List[object],object]

JWittmeyer (approved):

PR #169, opened 2025-05-21
Adds delta count methods

JWittmeyer:
main PR: https://github.com/code-kern-ai/refinery-gateway/pull/303

### DIFF
--- a/business_objects/record.py
+++ b/business_objects/record.py
@@ -485,6 +485,43 @@ def count(project_id: str) -> int:
     return session.query(Record).filter(Record.project_id == project_id).count()
 
 
+def count_missing_delta(project_id: str, attribute_id: str) -> int:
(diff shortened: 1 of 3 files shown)


business_objects/record.py
```diff
@@ -451,29 +451,47 @@ def get_full_record_data_for_id_group(
...
+    join_extension, where_add = "", ""
+    if only_missing:
+        if not embedding_id:
+            raise ValueError("embedding_id must be provided if only_missing is True")
+        join_extension, where_add = (
+            f"""
+        LEFT JOIN embedding_tensor et
+            ON et.project_id = r.project_id
+            AND et.record_id = r.id
+            AND et.project_id = '{project_id}' AND et.embedding_id = '{embedding_id}'
```
LennartSchmidtKern, line 478: better to sanitize embedding_id even it is coming from internal calls

PR #171, opened 2025-05-26
Templated steps

JWittmeyer:
main PR: https://github.com/code-kern-ai/cognition-gateway/pull/223

### DIFF
--- a/cognition_objects/step_templates.py
+++ b/cognition_objects/step_templates.py
@@ -145,10 +145,14 @@ def get_all_existing_steps_for_template_creation(org_id: str) -> Dict[str, Any]:
 
 
 def get_step_template_progress_text_lookup_for_strategy(
-    project_id: str, strategy_id: str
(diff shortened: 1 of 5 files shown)


cognition_objects/step_templates.py
```diff
@@ -0,0 +1,281 @@
...
+    ]
+
+    query = f"""
+    SELECT jsonb_object_agg(id,C)
+    FROM (
+        SELECT ss.config->>'templateId' id, COUNT(*)c
+        FROM cognition.strategy_step ss
+        INNER JOIN cognition.project p
+            ON ss.project_id = p.id
+        WHERE p.organization_id = '{organization_id}'
```
LennartSchmidtKern, line 42: prevent_sql_injection(...)

JWittmeyer: wasn't included since the orm query would fail prior to that but since that is a very specific usecase/exception and error prone in future changes best to add. => added

LennartSchmidtKern (approved):

PR #172, opened 2025-06-02
Deletion of multiple records

JWittmeyer:
main PR: https://github.com/code-kern-ai/refinery-gateway/pull/307

### DIFF
--- a/business_objects/record.py
+++ b/business_objects/record.py
@@ -773,7 +773,7 @@ def delete(project_id: str, record_id: str, with_commit: bool = False) -> None:
     general.flush_or_commit(with_commit)
 
 
-def delete_multiple(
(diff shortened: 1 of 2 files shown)


business_objects/record.py
```diff
@@ -773,6 +773,18 @@ def delete(project_id: str, record_id: str, with_commit: bool = False) -> None:
     general.flush_or_commit(with_commit)
 
 
+def delete_multiple(
```
andhreljaKern: we usually use the name `delete_many` for these kinds of functions, should we apply the same here?

JWittmeyer: changed

andhreljaKern (approved):

PR #187, opened 2025-09-16
Hotfix permission batch

LennartSchmidtKern:
Main PR:
- https://github.com/code-kern-ai/cognition-integration-provider/pull/41

### DIFF
--- a/cognition_objects/integration.py
+++ b/cognition_objects/integration.py
@@ -278,3 +278,20 @@ def get_sharepoint_permissions_by_integration_id(
         ON x.id = s.id
     """
     return session.execute(query).all()
+
+
+def get_distinct_item_ids_for_all_permissions(
+    integration_id: str,
+) -> List[str]:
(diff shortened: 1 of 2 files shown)


cognition_objects/integration.py
```diff
@@ -278,3 +278,20 @@ def get_sharepoint_permissions_by_integration_id(
...
+    )
+    query = f"""SELECT DISTINCT x.object_id
+    FROM (
+        SELECT json_array_elements_text(permissions) permission_id, MAX(object_id::TEXT) object_id
+        FROM integration.sharepoint
+        WHERE integration_id = '{integration_id}'
+        GROUP BY 1
+    ) x;"""
+    results = session.execute(query).all()
+    return [row[0] for row in results]
```
JWittmeyer: not sure if it can happen but maybe we should check if results exist so we dont iterate over None / NULL and return an empty []

JWittmeyer (approved):

PR #192, opened 2025-10-01
Added table release notifications and added business object

lumburovskalina:
Main PR: https://github.com/code-kern-ai/cognition-ui/pull/169

### DIFF
--- a/business_objects/release_notification.py
+++ b/business_objects/release_notification.py
@@ -28,3 +28,30 @@ def create(
     )
     general.add(message, with_commit)
     return message
+
+
+def update(
+    notification_id: str,
+    link: str,
+    config: str,
+    with_commit: bool = False,
+) -> ReleaseNotification:
+    release_notification = get(notification_id)
+
+    if release_notification is None:
+        return
+
+    if link is not None:
+        release_notification.link = link
+    if config is not None:
+        release_notification.config = config
+
+    general.flush_or_commit(with_commit)
+    return release_notification
+
+
+def delete(notification_id: str, with_commit: bool = False) -> None:
+    release_notification = get(notification_id)
+    general.delete(release_notification, with_commit)
+    general.flush_or_commit(with_commit)
+    return None
--- a/business_objects/release_notification.py
+++ b/business_objects/release_notification.py
@@ -14,7 +14,12 @@ def get(message_id: str) -> ReleaseNotification:
 
 
 def get_all() -> List[ReleaseNotification]:
-    return session.query(ReleaseNotification).filter().all()
+    return (
+        session.query(ReleaseNotification)
+        .filter()
+        .order_by(ReleaseNotification.created_at.desc())
+        .all()
+    )
 
 
 def create(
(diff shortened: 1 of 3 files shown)


business_objects/release_notification.py
```diff
@@ -0,0 +1,62 @@
...
+        session.query(ReleaseNotification)
+        .filter(ReleaseNotification.id == message_id)
+        .first()
+    )
+
+
+def get_all() -> List[ReleaseNotification]:
+    return (
+        session.query(ReleaseNotification)
+        .filter()
```
LennartSchmidtKern: empty filter

business_objects/release_notification.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def create(
+    link: str,
+    config: str,
+    created_by: str,
+    with_commit: bool = False,
+) -> ReleaseNotification:
+    message = ReleaseNotification(
+        link=link, config=config, created_by=created_by, created_at=datetime.utcnow()
```
LennartSchmidtKern: created_at should be set optionally and usually by db then in the model
created_at = Column(DateTime, default=sql.func.now())

business_objects/release_notification.py
```diff
@@ -0,0 +1,62 @@
+from typing import List
+from ..models import ReleaseNotification
+from . import general
+from ..session import session
+from datetime import datetime
+
+
+def get(message_id: str) -> ReleaseNotification:
```
LennartSchmidtKern: why do we refer in this file as message_id, when the object is ReleaseNotification?

business_objects/release_notification.py
LennartSchmidtKern: "message" vs. "notification" 
see aboove

business_objects/release_notification.py
```diff
@@ -0,0 +1,62 @@
...
+
+    general.flush_or_commit(with_commit)
+    return release_notification
+
+
+def delete(notification_id: str, with_commit: bool = False) -> None:
+    release_notification = get(notification_id)
+    general.delete(release_notification, with_commit)
+    general.flush_or_commit(with_commit)
+    return None
```
LennartSchmidtKern: no return instead of None

models.py
```diff
@@ -2490,3 +2490,17 @@ class AdminQueryMessageSummary(Base):
     deleted_messages_by_user = Column(Integer, default=0)
     deleted_messages_by_system = Column(Integer, default=0)
     incognito_messages = Column(Integer, default=0)
+
+
+class ReleaseNotification(Base):
+    __tablename__ = Tablenames.RELEASE_NOTIFICATION.value
+    __table_args__ = {"schema": "global"}
+    id = Column(Integer, primary_key=True, autoincrement=True)
```
LennartSchmidtKern, line 2498: Why do we use Integer as id?

lumburovskalina: because in the localstorage in the cognition ui we check by "bigger" ids if the notification was seen or not

business_objects/release_notification.py
lumburovskalina: changed

LennartSchmidtKern (approved):

PR #195, opened 2025-10-17
Incognito mode to conversations

lumburovskalina:
Main PR: https://github.com/code-kern-ai/cognition-ui/pull/175

### DIFF
--- a/cognition_objects/conversation.py
+++ b/cognition_objects/conversation.py
@@ -1,6 +1,6 @@
 from typing import Dict, List, Optional, Tuple, Any, Union
 
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from ..cognition_objects import message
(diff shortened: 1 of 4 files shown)


cognition_objects/conversation.py
```diff
@@ -70,6 +70,18 @@ def get_conversation_files_to_clean_up() -> List[Tuple[str, str, str]]:
     return general.execute_all(query)
 
 
+def get_conversations_older_than_24_hours() -> List[Tuple[str, str, str]]:
```
JWittmeyer: might be misleading since this is not just conversations oder than one hour but specifically. Get incognito conversations to delete. however i wouldn't even select them just run a deletion query

JWittmeyer (approved):

PR #199, opened 2025-10-28
Query for last chat messages

lumburovskalina:
Main PR: https://github.com/code-kern-ai/admin-dashboard/pull/91

### DIFF
--- a/cognition_objects/integration.py
+++ b/cognition_objects/integration.py
@@ -334,69 +334,93 @@ def get_distinct_item_ids_for_all_permissions(
 def get_last_integrations_tasks() -> List[Dict[str, Any]]:
     query = f"""
     WITH embedding_agg AS (
-    SELECT 
-        project_id,
-        jsonb_agg(
-            jsonb_build_object(
-                'createdBy', e.created_by,
-                'finishedAt', e.finished_at,
-                'id', e.id,
-                'name', e.name,
-                'startedAt', e.started_at,
-                'state', e.state
-            )
-        ) AS embeddings
-    FROM embedding e
-    GROUP BY project_id
+        SELECT 
+            project_id,
+            jsonb_agg(
+                jsonb_build_object(
+                    'createdBy', e.created_by,
+                    'finishedAt', e.finished_at,
+                    'id', e.id,
+                    'name', e.name,
+                    'startedAt', e.started_at,
+                    'state', e.state
+                ) ORDER BY e.started_at DESC
+            ) AS embeddings
+        FROM embedding e
+        GROUP BY project_id
     ),
 
     attribute_agg AS (
-    SELECT 
-        project_id,
-        jsonb_agg(
-            jsonb_build_object(
-                'dataType', a.data_type,
-                'finishedAt', a.finished_at,
-                'id', a.id,
-                'name', a.name,
-                'startedAt', a.started_at,
-                'state', a.state
-            )
-        ) AS attributes
-    FROM "attribute" a
-    GROUP BY project_id
+        SELECT 
+            project_id,
+            jsonb_agg(
+                jsonb_build_object(
+                    'dataType', a.data_type,
+                    'finishedAt', a.finished_at,
+                    'id', a.id,
+                    'name', a.name,
+                    'startedAt', a.started_at,
+                    'state', a.state
+                ) ORDER BY a.started_at DESC
+            ) AS attributes
+        FROM "attribute" a
+        GROUP BY project_id
     ),
 
     record_tokenization_task_agg AS (
-    SELECT 
-        project_id,
-        jsonb_agg(
+        SELECT 
+            project_id,
+            jsonb_agg(
+                jsonb_build_object(
+                    'finishedAt', rtt.finished_at,
+                    'id', rtt.id,
+                    'startedAt', rtt.started_at,
+                    'state', rtt.state,
+                    'type', rtt.type
+                ) ORDER BY rtt.started_at DESC
+            ) AS record_tokenization_tasks
+        FROM record_tokenization_task rtt
+        GROUP BY project_id
+    ),
+
+    integration_data AS (
+        SELECT 
+            i.id AS integration_id,
+            i.name AS integration_name,
+            i.error_message,
+            i.started_at,
+            i.finished_at,
+            i.state,
+            i.organization_id,
             jsonb_build_object(
-                'finishedAt', rtt.finished_at,
-                'id', rtt.id,
-                'startedAt', rtt.started_at,
-                'state', rtt.state,
-                'type', rtt.type
-            )
-        ) AS record_tokenization_tasks
-    FROM record_tokenization_task rtt
-    GROUP BY project_id
+                'embeddings', coalesce(ea.embeddings, '[]'::jsonb),
+                'attributes', coalesce(aa.attributes, '[]'::jsonb),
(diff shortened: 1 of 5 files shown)


cognition_objects/integration.py
```diff
@@ -329,3 +330,171 @@ def get_distinct_item_ids_for_all_permissions(
...
+                    )
+                )
+                FROM embeddings_by_state ebs
+                WHERE ebs.project_id = e.project_id
+            ) AS embeddings_by_state
+
+        FROM embedding e
+        LEFT JOIN project p ON p.id = e.project_id
+        GROUP BY e.project_id
+    ),
```
JWittmeyer: why do we need to join with project?
Also if we just want the object aggregated embeddings this could be one query instead of the embeddings_by_state & embedding_agg  by wrapping it in another select

cognition_objects/integration.py
```diff
@@ -329,3 +330,171 @@ def get_distinct_item_ids_for_all_permissions(
...
+                )
+                FROM attribute_by_state abs
+                WHERE abs.project_id = a.project_id
+            ) AS attributes_by_state
+
+        FROM attribute a
+        LEFT JOIN project p ON p.id = a.project_id
+        WHERE a.state NOT IN ('UPLOADED','AUTOMATICALLY_CREATED')
+        GROUP BY a.project_id
+    ),
```
JWittmeyer: same as above

cognition_objects/integration.py
```diff
@@ -329,3 +330,171 @@ def get_distinct_item_ids_for_all_permissions(
...
+                    )
+                )
+                FROM record_tokenization_tasks_by_state rtts
+                WHERE rtts.project_id = rtt.project_id
+            ) AS record_tokenization_tasks_by_state
+
+        FROM record_tokenization_task rtt
+        LEFT JOIN project p ON p.id = rtt.project_id
+        GROUP BY rtt.project_id
+    ),
```
JWittmeyer, line 428: same as above

cognition_objects/integration.py
```diff
@@ -329,3 +330,171 @@ def get_distinct_item_ids_for_all_permissions(
...
+            i.type,
+            jsonb_build_object(
+                'embeddingsByState', coalesce(ea.embeddings_by_state, '[]'::jsonb),
+                'attributesByState', coalesce(aa.attributes_by_state, '[]'::jsonb),
+                'recordTokenizationTasksByState', coalesce(rtt.record_tokenization_tasks_by_state, '[]'::jsonb)
+            ) AS full_data
+        FROM cognition.integration i
+        LEFT JOIN embedding_agg ea ON ea.project_id = i.project_id
+        LEFT JOIN attribute_agg aa ON aa.project_id = i.project_id
+        LEFT JOIN record_tokenization_task_agg rtt ON rtt.project_id = i.project_id
```
JWittmeyer: we usually write the on in the next line so e.g.
LEFT JOIN embedding_agg ea
    ON ea.project_id = i.project_id

cognition_objects/integration.py
```diff
@@ -329,3 +330,171 @@ def get_distinct_item_ids_for_all_permissions(
...
+        i.started_at,
+        i.finished_at,
+        i.state,
+        i.full_data,
+        i.created_by,
+        i.type,
+        p.name AS project_name
+    FROM organization o
+    JOIN integration_data i ON i.organization_id = o.id
+    LEFT JOIN project p ON p.id = i.project_id
```
JWittmeyer: afaik doesn't need to be a left join since prject should always exist for an integration

Also we could build this directly in the previousintegration data by joining there with org & project

cognition_objects/markdown_file.py
```diff
@@ -218,3 +218,45 @@ def delete_many(org_id: str, md_file_ids: List[str], with_commit: bool = True) -
...
+            ROW_NUMBER() OVER (
+                PARTITION BY md.organization_id, md.id
+                ORDER BY mf.created_at DESC
+            ) AS rn
+        FROM cognition.markdown_file mf
+            JOIN cognition.markdown_dataset md ON md.id = mf.dataset_id
+            JOIN organization o ON o.id = md.organization_id
+        WHERE 
+            mf.created_at >= '{created_at_from}'
+            AND mf.state IN ({', '.join(f"'{state}'" for state in states)})
```
JWittmeyer: could potentially fail if states is empty list so maybe an early return for this since without any state it would be empty anyway

cognition_objects/markdown_file.py
```diff
@@ -218,3 +218,45 @@ def delete_many(org_id: str, md_file_ids: List[str], with_commit: bool = True) -
...
     general.flush_or_commit(with_commit)
+
+
+def get_last_etl_tasks(
+    states: List[str],
+    created_at_from: str,
+    created_at_to: Optional[str] = None,
+) -> List[Any]:
+
+    states = prevent_sql_injection(states, isinstance(states, list))
```
JWittmeyer: afaik we need to prevent every string with a list comprehension not the whole list

cognition_objects/message.py
```diff
@@ -605,3 +607,48 @@ def get_count_by_project_id(project_id: str) -> int:
         )
         .count()
     )
+
+
+def get_last_chat_messages(
+    message_type: str,
```
JWittmeyer: we could give it the type in the request body directly so fastapi parses it and we can compare without .value

cognition_objects/strategy.py
```diff
@@ -107,3 +109,76 @@ def delete_all_by_project_id(project_id: str, with_commit: bool = True) -> None:
...
     general.flush_or_commit(with_commit)
+
+
+def get_strategies_info(
+    step_types: List[str],
+    created_at_from: str,
+    created_at_to: Optional[str] = None,
+) -> List[Any]:
+
+    step_types = prevent_sql_injection(step_types, isinstance(step_types, list))
```
JWittmeyer: see above

cognition_objects/strategy.py
```diff
@@ -107,3 +109,76 @@ def delete_all_by_project_id(project_id: str, with_commit: bool = True) -> None:
...
+    if created_at_to:
+        created_at_to = prevent_sql_injection(
+            created_at_to, isinstance(created_at_to, str)
+        )
+    created_at_to_filter = ""
+
+    if created_at_to:
+        created_at_to_filter = f"AND ss.created_at <= '{created_at_to}'"
+
+    step_types_sql = ", ".join([f"'{st}'" for st in step_types])
```
JWittmeyer, line 136: could be faulty for empty lists so early return recommended

cognition_objects/strategy.py
```diff
@@ -107,3 +109,76 @@ def delete_all_by_project_id(project_id: str, with_commit: bool = True) -> None:
...
+        {created_at_to_filter}
+    )
+    SELECT strategy_id, strategy_name, step_id, created_by, created_at, step_name, step_type, project_name,project_id, organization_name,organization_id,
+        CASE
+            WHEN step_type = '{StrategyStepType.TEMPLATED.value}'
+                AND EXISTS (
+                    SELECT 1
+                    FROM jsonb_array_elements(template_config->'steps') t
+                    WHERE t->>'stepType' IN ({step_types_sql})
+                )
```
JWittmeyer: exists filter are usually slow. and they are run multiple times in the query. maybe we can preselect the step type  (single value) and the template step types (array) in the qith query to then in a sourrounding query filter on whats needed

JWittmeyer (approved):

PR #210, opened 2026-01-12
Postrelease Hotfixes

LennartSchmidtKern:
Main PR:
- https://github.com/code-kern-ai/cognition-ui/pull/190


global_objects/inbox_mail.py
```diff
@@ -386,3 +386,59 @@ def update_thread_progress(
...
+) -> None:
+    thread_entity = get_inbox_mail_thread_by_id(thread_id)
+    project_id = thread_entity.meta_data.get("projectId")
+    if not project_id:
+        return
+
+    threads = (
+        session.query(InboxMailThread)
+        .filter(
+            InboxMailThread.is_admin_support_thread == True,
```
JWittmeyer: maybe we should also filter that the email isn't read already so it doesn't collect that much

JWittmeyer (approved):

PR #212, opened 2026-01-19
Webpage Integration

LennartSchmidtKern:
Main PR:
- https://github.com/code-kern-ai/cognition-integration-provider/pull/71


business_objects/record.py
```diff
@@ -1006,3 +1007,16 @@ def get_record_data_by_sanitized_where(
     """
     data = general.execute_all(query)
     return [row[0] for row in data] if data else []
+
+
+def get_record_by_attribute_value(
```
andhreljaKern: unused

integration_objects/manager.py
```diff
@@ -32,17 +37,67 @@ def get(
...
+) -> int:
+    IntegrationModel = integration_model(integration=integration)
+    records = (
+        session.query(IntegrationModel)
+        .filter(
+            IntegrationModel.integration_id == integration.id,
+        )
+        .all()
+    )
+    filtered_records = [
```
andhreljaKern: use a single query instead of list comprehension for filtering

integration_objects/manager.py
```diff
@@ -32,17 +37,67 @@ def get(
...
+        x for x in records if re.search(r"#\d$", getattr(x, by, x.source) or "")
+    ]
+
+    for record in filtered_records:
```
andhreljaKern: unused

etl_utils.py
```diff
@@ -169,7 +170,65 @@ def get_full_config_and_tokenizer_from_config_id(
...
+                "use_cache": False,
+                "fallback": None,
+            },
+        },
+        {
+            "llm_config": integration.llm_config,
+            "task_type": enums.CognitionMarkdownFileState.SPLITTING.value,
+            "task_config": {
+                "use_cache": False,
+                "strategy": enums.ETLSplitStrategy.CHUNK.value,
```
andhreljaKern, line 196: expose strategy selection to UI?

etl_utils.py
```diff
@@ -187,7 +246,7 @@ def get_full_config_for_integration(
             "task_type": enums.CognitionMarkdownFileState.SPLITTING.value,
             "task_config": {
                 "use_cache": False,
-                "strategy": enums.ETLSplitStrategy.SHRINK.value,
+                "strategy": enums.ETLSplitStrategy.CHUNK.value,
```
andhreljaKern, line 248: should be configurable, at the least not hardcoded to CHUNK

cognition_objects/integration.py
JWittmeyer, line 1: 🔴 **CRITICAL**: execution_finished() returns False when integration.state is FINISHED or FAILED. New logic only returns True when state is ETL_PROCESSING and no active ETL tasks exist. When state is FINISHED/FAILED, callers will get False (execution not finished), which is incorrect and can break workflows that rely on "finished" meaning state in FINISHED_STATES.

```suggestion
def execution_finished(id: str) -> bool:
    integration = get_by_id(id)
    if not integration:
        return True
    if integration.state in FINISHED_STATES:
        return True
    if integration.state not in ETL_PROCESSING_STATES:
        return False
    return not exists_active_etl_tasks(id)
```

JWittmeyer, line 1: 🔴 **CRITICAL**: has_scheduled_etl_tasks() can raise AttributeError when get_by_id() returns None (integration not found). No None check before accessing integration.state.

```suggestion
def has_scheduled_etl_tasks(integration_id: str) -> bool:
    integration = get_by_id(integration_id)
    if not integration:
        return False
    return integration.state not in FINISHED_STATES
```

global_objects/etl_task.py
JWittmeyer, line 1: 🔴 **CRITICAL**: get_or_create_integration_etl_task() no longer returns an existing EtlTask when record.etl_task_id is set. The early return that looked up and returned the existing task was removed, so every call creates a new EtlTask and can create duplicate tasks for the same record.

integration_objects/manager.py
JWittmeyer, line 1: 🔴 **CRITICAL**: get_record_scope() returns None when last_record is None or etl_task is None (bare 'return' with no value). Return type is IntegrationRecordScope; callers may expect a scope string/value and get TypeError or incorrect behavior.

```suggestion
if last_record is None:
    return IntegrationRecordScope.ROOT  # or ALL, per product intent
...
if etl_task is None:
    return IntegrationRecordScope.ROOT
```

JWittmeyer, line 1: 🔴 **CRITICAL**: Scope filter for ROOT uses record_identifier.like("%[^#]%"). In SQL LIKE, [^#] is not a character class (that is regex syntax); it matches the literal four characters [^#]. Root records (no #digit suffix) will not be selected correctly. Use a DB regex operator or NOT LIKE '%#_' with proper escaping for "does not end with #digit".

enums.py
JWittmeyer, line 1: 🔴 **CRITICAL**: get_default_extractor() has no branch for ETLFileType.TSV. CSV and JSON were added; TSV falls through and raises ValueError("No default extractor for given file type {file_type}").

```suggestion
elif file_type == ETLFileType.TSV:
    return ETLExtractorTsv.LANGCHAIN
```

integration_objects/manager.py
JWittmeyer, line 1: 🟡 **SUGGESTION**: count() uses len(session.query(...).all()) instead of .count(), loading all rows into memory. For large integrations this hurts performance; use .count() for existence/count.

```suggestion
return (
    session.query(IntegrationModel)
    .filter(...)
    .filter(record_identifier.op("regexp")(r"#\d$"))
    .count()
) > 0
```

JWittmeyer, line 1: 🟡 **SUGGESTION**: get_all_by_etl_task() calls etl_task.meta_data.get() without checking if meta_data is None. If meta_data is None, AttributeError will be raised.

```suggestion
integration_id = (etl_task.meta_data or {}).get("integration_id")
```

JWittmeyer, line 1: 🟡 **SUGGESTION**: set_refinery_synced_by_record_ids() does not guard against empty record_ids. Filtering with id.in_([]) can be invalid or unexpected in some databases; early return when record_ids is empty.

```suggestion
if not record_ids:
    return
```

JWittmeyer, line 1: 🔵 **NOTE**: duplicate() uses except Exception with print() and re-raise. Review policy prefers proper logging over print for errors; consider logging.exception() before raise.

JWittmeyer, line 1: ❓ **QUESTION**: create() no longer accepts or sets running_id on the new integration record. If running_id is required elsewhere (e.g. get_running_ids, uniqueness), was a default or follow-up assignment added on the model or in callers?

JWittmeyer, line 1: ❓ **QUESTION**: update() now sets etl_task_id whenever it is not None (previously only when integration_record.etl_task_id was None). Was overwriting an existing etl_task_id intentional?

JWittmeyer:
🤖 Reviewed as part of PR group anchored at [code-kern-ai/cognition-integration-provider#71](https://github.com/code-kern-ai/cognition-integration-provider/pull/71). See the main PR for the full summary.

JWittmeyer (approved):

PR #216, opened 2026-02-03
Cursor files submodules model

lumburovskalina:
Main PR: https://github.com/code-kern-ai/refinery-gateway/pull/377


.cursor/rules/guidelines.mdc
```diff
@@ -0,0 +1,71 @@
...
+
+### Common Patterns
+
+- Entity models inherit from `Base` (SQLAlchemy declarative_base)
+- Business objects provide `get()`, `create()`, `update()`, `delete()` functions
+- Use `with_commit` parameter to control transaction boundaries
+- Foreign keys use `ondelete` cascades appropriately
+- UUIDs are used for primary keys
+- Timestamps use `created_at` with `default=sql.func.now()`
+- **After schema changes**: Run `./db commit "message"` then `./db migrate`
```
LennartSchmidtKern: Maybe this will make problems, because its stated in the refinery-gateway already. Remove here or cross reference? Not sure about best practice. From the submodules perspective, it does not know about refinery-gateway if its started from somewhere else e. g. cognition-gateway

.cursor/rules/business-objects.mdc
```diff
@@ -0,0 +1,452 @@
...
+
+### File Naming
+
+- One file per entity
+- File name matches entity name in `snake_case`
+- Example: `user.py` for `User` entity, `conversation.py` for `Conversation` entity
+
+## Standard CRUD Functions
+
+Every business object module should provide these standard functions:
```
LennartSchmidtKern: Could be confusing for the agent, because the examples are all using project_ids, but it could be independent of projects featurewise

.cursor/rules/business-objects.mdc
```diff
@@ -0,0 +1,452 @@
...
+        created_at=created_at,
+        # ... other fields
+    )
+    
+    general.add(entity, with_commit)
+    return entity
+```
+
+**Rules:**
+- Always accept `with_commit` parameter (default `False`)
```
LennartSchmidtKern: not true for cognition I think

.cursor/rules/business-objects.mdc
```diff
@@ -0,0 +1,452 @@
...
+    
+    flush_or_commit(with_commit)
+    return entity
+```
+
+**Rules:**
+- All update parameters should be `Optional`
+- Use `None` checks to only update provided fields
+- Use `flag_modified()` for JSON fields
+- Default `with_commit=True` for updates (different from create)
```
LennartSchmidtKern: not true for refinery, maybe explain this in general somewhere

.cursor/rules/business-objects.mdc
```diff
@@ -0,0 +1,452 @@
...
+- Use for bulk deletion operations
+- Default `with_commit=False` to allow batching
+- Consider performance implications for large batches
+
+## Transaction Management
+
+### `with_commit` Parameter
+
+- **`False`**: Flush to database but don't commit (default for create/delete)
+- **`True`**: Commit transaction immediately (default for update)
```
LennartSchmidtKern, line 177: see above, probably here

.cursor/rules/database-operations.mdc
```diff
@@ -0,0 +1,333 @@
...
+from ..business_objects import general
+
+entity = general.refresh(entity)
+```
+
+## Error Handling
+
+### Session Rollback on Error
+
+Always rollback on exceptions:
```
LennartSchmidtKern: thats not true, because its handled in our session remove and refresh function

.cursor/rules/database-operations.mdc
```diff
@@ -0,0 +1,333 @@
...
+
+### Rollback
+
+```python
+from ..business_objects import general
+
+general.rollback()
+```
+
+**Use rollback** when errors occur to maintain data consistency.
```
LennartSchmidtKern: see below

.cursor/rules/database-operations.mdc
```diff
@@ -0,0 +1,333 @@
...
+from ..session import request_id_ctx_var, get_request_id
+
+# Get current request ID
+request_id = get_request_id()
+```
+
+### Session Lifecycle
+
+- Sessions are automatically scoped per request
+- Use `general.get_ctx_token()` to create a new session context
```
LennartSchmidtKern: maybe adding that we normally do not need to do that, only for threads in very specific cases and if its not handled in the repos middleware

LennartSchmidtKern: We also need some info about daemon with db session probably

.cursor/rules/entities.mdc
```diff
@@ -0,0 +1,369 @@
...
+# 1. Add new column to models.py
+class Project(Base):
+    __tablename__ = Tablenames.PROJECT.value
+    # ... existing columns ...
+    new_feature_flag = Column(Boolean, default=False)  # New column
+```
+
+```bash
+# 2. Create migration
+./db commit "Add new_feature_flag to Project table"
```
LennartSchmidtKern, line 358: Same already mentioned in relation to refinery-gateway context

.cursor/rules/guidelines.mdc
lumburovskalina: i will leave it only in the refinery-gateway since it should be only done from there, it the other repos it is stated that it should be only done from refinery-gateway

.cursor/rules/business-objects.mdc
lumburovskalina: removed, since it is explained below

LennartSchmidtKern (approved):

PR #240, opened 2026-04-15
New helper page for recent negative feedback

lumburovskalina:
Main PR: https://github.com/code-kern-ai/admin-dashboard/pull/108

### DIFF
--- a/cognition_objects/message.py
+++ b/cognition_objects/message.py
@@ -660,3 +660,63 @@ def get_last_chat_messages(
     """
 
     return general.execute_all(query)
+
+
+def get_last_negative_feedback_per_org(
+    created_at_from: str,
+    created_at_to: Optional[str] = None,
+    cross_selling_filter: Optional[str] = None,
+    feedback_category: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


JWittmeyer:
🤖 Reviewed as part of PR group anchored at [code-kern-ai/admin-dashboard#108](https://github.com/code-kern-ai/admin-dashboard/pull/108). See the main PR for the full summary.

cognition_objects/message.py
```diff
@@ -660,3 +660,63 @@ def get_last_chat_messages(
     """
 
     return general.execute_all(query)
+
+
+def get_last_negative_feedback_per_org(
```
JWittmeyer, line 665: name might be misleading when we have last negative but provide a time filter. or is this last 5 in that area? (with the rn filter yes not sure if understandable in, if i filter an area i might want all not just the last 5)

JWittmeyer (approved):

cognition_objects/message.py
lumburovskalina: discussed and resolved

PR #242, opened 2026-04-20
Tables improvements

lumburovskalina:
Main PR: https://github.com/code-kern-ai/cognition-ui/pull/214

### DIFF
--- a/util.py
+++ b/util.py
@@ -218,6 +218,13 @@ def to_frontend_obj(
             )
             for x in value
         ]
+    elif isinstance(value, Row) or isinstance(value, Base):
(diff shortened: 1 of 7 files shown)


util.py
```diff
@@ -218,6 +218,13 @@ def to_frontend_obj(
             )
             for x in value
         ]
+    elif isinstance(value, Row) or isinstance(value, Base):
```
andhreljaKern: could potentially cause an infinite loop because `sql_alchemy_to_dict` calls `to_frontend_obj`

andhreljaKern (approved):

